PR #9, opened 2025-07-14
libsdl2-cocoa: fix header issue that breaks build on Tiger

ForestExpertise:
I have tested this portfile and libsdl2-cocoa builds successfully on Tiger with it. Thank you for the fix! I thought it should perhaps be added to the portfile if it does not break anything on Leopard or Snow Leopard.


barracuda156:
@ForestExpertise Thank you!

PR #10, opened 2025-07-23
Amule: Change boost version so all dependencies can build on Tiger

ForestExpertise:
Per kencu here: https://trac.macports.org/ticket/71753
and also my own testing, boost178 works on Tiger ppc while earlier boost versions do not. While this ticket is about boost171, I suspect that a similar issue afflicts boost176. I was able to build all dependencies for amule on ppc Tiger after this portfile change. Another option would be to fix a boost port and then use that, and I am happy to provide logs and testing, but I think it might be easier to use boost178 on Tiger when relevant. I would love to hear your opinion on which route makes sense, as you have far more experience than my less than three months with Macports.

### DIFF
--- a/net/amule/Portfile
+++ b/net/amule/Portfile
@@ -10,7 +10,7 @@ PortGroup           wxWidgets 1.0
 if {${os.platform} ne "darwin" || ${os.major} > 9} {
     boost.version   1.87
 } else {
-    boost.version   1.76
+    boost.version   1.78
 }
 
 github.setup        amule-project amule 9ceeaa68b9727fa38efd9ddcf774b20d39d5a200


net/amule/Portfile
```diff
@@ -10,7 +10,7 @@ PortGroup           wxWidgets 1.0
 if {${os.platform} ne "darwin" || ${os.major} > 9} {
     boost.version   1.87
 } else {
-    boost.version   1.76
+    boost.version   1.78
```
barracuda156, line 13: This should be a draft of a PR until it is actually confirmed that `amule` builds against boost 1.78 (it is not automatically guaranteed). I also think that 1.76 should work fine, but I won’t object switching to 1.78. If the latter is done, then revision must be increased for relevant systems, because it could have been installed by someone on 10.5 with boost 1.76. In this case it will be easier to add `incr revision` inside this block (i.e. above or below `boost.version 1.78` line).

ForestExpertise: Thanks for letting me know. I will close this PR until such time as I can get amule to build.

barracuda156: Could you say what specifically fails?

Issue #2, opened 2024-12-26
ghc 7.6.3 fails to build with gcc-ppc-bootstrap 7.0.4 due to libffi bug affecting 10.6.8


barracuda156:
This is specific to 10.6.8:
```
--->  Configuring ghc-bootstrap
Executing:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_lang_ghc-bootstrap/ghc-bootstrap/work/ghc-7.6.3" && ./configure --prefix=/opt/local/share/ghc-bootstrap --with-ghc=/opt/local/share/ghc-ppc-bootstrap/bin/ghc --with-iconv-includes=/opt/local/include --with-iconv-libraries=/opt/local/lib --with-gmp-includes=/opt/local/include --with-gmp-libraries=/opt/local/lib --with-macosx-deployment-target=10.6 --with-gcc=/usr/bin/gcc-4.2 
checking for gfind... no
checking for find... /usr/bin/find
checking for sort... /usr/bin/sort
checking version of ghc... 7.0.4
checking build system type... powerpc-apple-darwin10.8.0
checking host system type... powerpc-apple-darwin10.8.0
checking target system type... powerpc-apple-darwin10.8.0
HOST: powerpc-apple-darwin10.8.0
Build platform inferred as: powerpc-apple-darwin
Host platform inferred as: powerpc-apple-darwin
Target platform inferred as: powerpc-apple-darwin
GHC build  : powerpc-apple-darwin
GHC host   : powerpc-apple-darwin
GHC target : powerpc-apple-darwin
configure: Building in-tree ghc-pwd
ld: in /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/libHSffi.a(darwin.o), unsupported encoding in FDE
collect2: ld returned 1 exit status
configure: error: Building ghc-pwd failed
[...]

PR #13, opened 2025-08-13
Gawk: Add Tiger fallback, avoid patch try #2

ForestExpertise:
This is my revised pull request in accordance with feedback. I have not had the chance to test it yet. If you like, I can test and confirm this works in a comment before proceeding.


barracuda156:
@ForestExpertise I did confirm the build on Tiger, but if something does not work as it should, we can always restore a fallback to an older version, just let me know.

PR #22, opened 2025-11-10
ld64-97: fix rpath on Tiger

ForestExpertise:
I found some brilliant patches from kencu that fix a lot of problems here: https://trac.macports.org/ticket/57975
These patches allow many things to build and work that would not otherwise on tiger (such as the gstreamer plugins used in lollypop). I have been testing it for a couple of weeks. May we create a variant so others can take advantage of these excellent patches?

### DIFF
--- a/devel/ld64/Portfile
+++ b/devel/ld64/Portfile
@@ -78,6 +78,17 @@ subport ld64-97 {
         ld64-97-no-ppc-thread_status.patch \
         ld64-97-configured_architectures.patch
 
+    # from https://trac.macports.org/ticket/57975
+    platform darwin 8 {
+        variant tigerrpath {
+            patchfiles-append \
+                ld64-97-no-error-rpath-tiger.diff \
+                ld64-97-dont-output-rpath-atom.diff
+        }
+        default_variants +tigerrpath
+    }
+            
+
     patch_sites         https://distfiles.macports.org/${name}/
     checksums-append    ld64-97-standalone-libunwind-headers.patch \
                         rmd160  f6da71e097aa61b1055b3fdc12cd39aafed5f492 \


barracuda156:
@ForestExpertise Let’s allow a few days if we get response from Ken, if not, then I can merge this as a variant.

P. S. If you wanna give credit or refer to someone, it is better to use either GH ID, or full name, or commit reference, i.e. something clearly identifiable for any third person.

kencu:
I believe those patches will allow some Tiger builds to succeed whereas they would otherwise fail, and are unlikely to cause unforeseen errors. The non-functional rpath commands may cause some things to fail, for example running tests during meson builds. But likely a step in the right direction for Tiger.

A more robust fix would be to modify the code so that instead of adding an rpath search path, the code added a full normal library path. 

Perhaps that might be done by modifying this patch:

```
$ cat ld64-97-dont-output-rpath-atom.diff
diff --git src/ld/MachOWriterExecutable.hpp.orig src/ld/MachOWriterExecutable.hpp
index 630328a..ee21fe2 100644
--- src/ld/MachOWriterExecutable.hpp.orig
+++ src/ld/MachOWriterExecutable.hpp
@@ -3142,7 +3142,7 @@ Writer<A>::Writer(const char* path, Options& options, std::vector<ExecutableFile
 	
 	// add any rpath load commands
 	for(std::vector<const char*>::const_iterator it=fOptions.rpaths().begin(); it != fOptions.rpaths().end(); ++it) {
-		fWriterSynthesizedAtoms.push_back(new RPathLoadCommandsAtom<A>(*this, *it));
+//KEN		fWriterSynthesizedAtoms.push_back(new RPathLoadCommandsAtom<A>(*this, *it));
 	}
 	
 	// set up fSlideable
```

to add in a full -L load path somehow, instead of outputting nothing.  

If that could be done, and if the place somewhere in the ld64 code where the the library name is added and have ```@rpath/``` changed to ```@executable_path/``` or perhaps  ```@loader_path/```  then the final product might even work as though ```@rpath``` was supported.

devel/ld64/Portfile
```diff
@@ -78,6 +78,15 @@ subport ld64-97 {
         ld64-97-no-ppc-thread_status.patch \
         ld64-97-configured_architectures.patch
 
+    variant tigerrpath {
```
barracuda156: The logic should be reverse here. We do not need this variant outside of 10.4, and certainly a variant which just does nothing.

ForestExpertise:
Keeping the original as an option also has some use for debugging - one can note which ports fail with the vanilla version but succeed with the tiger path version as ones that need rpath fixes. Though if one is good at reading logs, perhaps that benefit is minimal (I'm not though).

barracuda156:
Thank you!

PR #23, opened 2025-11-15
xapian-core: fix build on Tiger

ForestExpertise:
As discussed on the Macrumors forum, I am including my patch that allows xapian-core to build on Tiger. Let me know if if is ready to merge.

### DIFF
--- a/devel/xapian-core/files/tiger-needs-header.diff
+++ b/devel/xapian-core/files/tiger-needs-header.diff
@@ -1,13 +1,10 @@
---- backends/glass/glass_version.h
-+++ backends/glass/glass_version.h
-@@ -22,9 +22,10 @@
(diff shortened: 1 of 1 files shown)


devel/xapian-core/Portfile
```diff
@@ -33,6 +33,10 @@ if {${subport} eq ${name}} {
                                 sha256  c55c9bc8613ad3ec2c218eafca088c218ab7cddcba7ef08f3af0e542f4e521bc \
                                 size    3250704
 
+    darwin 8 {
```
barracuda156: It should be `platform darwin 8` (at least that was always the syntax).

barracuda156 (changes requested):

barracuda156:
Thank you!

PR #24, opened 2025-11-16
ninja: add Tiger fallback

ForestExpertise:
The following builds for me on 10.4.11. Let me know if it can be merged or if changes are recommended.

### DIFF
--- a/devel/ninja/Portfile
+++ b/devel/ninja/Portfile
@@ -35,7 +35,6 @@ homepage            https://ninja-build.org
 checksums           rmd160  6765fd1becaccc480b857bdf7953405dfdd43ff3 \
                     sha256  f0055ad0369bf2e372955ba55128d000cfcc21777057806015b45e4accbebf23 \
                     size    292098
-
 github.tarball_from archive
 checksums-prepend   ${distfiles}
 set gtest_version   1.16.0


devel/ninja/Portfile
```diff
@@ -3,6 +3,11 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
+platform darwin 8 {
+    PortGroup       legacysupport 1.1
+    legacysupport.newest_darwin_requires_legacy 10
```
barracuda156: This should be done differently.
```
PortSystem          1.0
PortGroup           github 1.0
PortGroup           legacysupport 1.1

legacysupport.newest_darwin_requires_legacy 8
```
(Assuming it is needed for Tiger.)

Could you share a log what actually fails with the current Ninja? (Sorry, if I forgot, but in any case it is rather be referenced in the PR, so that the reason for a fallback is clear.)

devel/ninja/Portfile
```diff
@@ -32,9 +37,20 @@ homepage            https://ninja-build.org
...
+    revision            1
+    checksums           ninja-1.12.1.tar.gz \
+                        rmd160  edb99cb67641947a6cdde89300a326de69df77f4 \
+                        sha256  821bdff48a3f683bc4bb3b6f0b5fe7b2d647cf65d52aeb63328c91a6c6df285a \
+}
+
 github.tarball_from archive
 checksums-prepend   ${distfiles}
 set gtest_version   1.16.0
+platform darwin 8 {
```
barracuda156: It will be easier to maintain (both for 10.4 and for 10.5+) if all 10.4-specific code goes into a single block somewhere below. If you prefer, I can handle that, you can just test the build then.

devel/ninja/Portfile
```diff
@@ -55,6 +64,26 @@ patchfiles          patch-configure.py-bootstrap-only.diff \
 # https://github.com/ninja-build/ninja/issues/2645
 patchfiles-append   8f2d7488ff69c68398112ded57496fa26081f4d7.patch
 
+platform darwin 8 {
+    checksums-delete    ${gtest_distfile} \
```
barracuda156: I am genuinely surprised this can be needed. Normally a variable overrides its earlier instances, unless `-append` is used (there are some special cases for convenience).

Does this work (assuming these are all patches needed)?
```
platform darwin 8 {
    set ninja_fallback  1.12.1
    set gtest_version   1.14.0
    github.setup        ninja-build ninja ${ninja_fallback} v
    revision            1
    checksums           ${name}-${ninja_fallback}.tar.gz \
                        rmd160  edb99cb67641947a6cdde89300a326de69df77f4 \
                        sha256  821bdff48a3f683bc4bb3b6f0b5fe7b2d647cf65d52aeb63328c91a6c6df285a \
                        size    240483 \
                        googletest-${gtest_version}.tar.gz \
                        rmd160  54f7bd68b80152528dfc13d9291b247c0353c2a4 \
                        sha256  8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7 \
                        size    867764

    patchfiles          patch-tiger-no-posix-spawn.diff \
                        patch-tiger-sysconf.diff \
                        tests.patch
}
```
And place this below, please (provided the build still works, of course). Perhaps where the old block was.

I wanna avoid trying to figure out every time which code block is for which system and what is their effect on each LOL. If this way is inconvenient for you to maintain, we can make a tree here and add its sources specifically for Tiger. But in either case clarity is among priorities, IMO. (Consider someone else will be reading this, which is unfamiliar with MacPorts.)

ForestExpertise: It does not. I get the
Error: No checksum set for ninja-1.13.1.tar.gz
---> Checksumming googletest-1.16.0.tar.gz
Error: No checksum set for googletest-1.16.0.tar.gz

I'm in agreement that having just one block for darwin 8 would be best, the Portfile is just surprisingly resistant to that. I'll keep playing around with it - I got it down to two blocks this week, maybe I'll figure out how to get it down to one. Feel free to throw out another suggestion for me to test if it is not much trouble.

ForestExpertise:
Good catch on the whitespace. I deleted it and it should now merge cleanly

PR #27, opened 2025-11-28
llvm-powerpc: add patch to fix build for Tiger

ForestExpertise:
I was pleasantly surprised that llvm-powerpc only needs one additional patch to build for Tiger. @catap did a lot of hard work almost making clang-11-bootstrap work on Tiger, and I took the minimum needed from his modification of Jeremy Huddleston Sequoia's patch that can be found here: https://github.com/macports/macports-ports/compare/master...catap:macports-ports:clang-11-bootstrap-work
The relevant ticket about the issue is found here: https://trac.macports.org/ticket/67348
But I have also attached a log, as llvm-powerpc is not exactly the same as clang-11-bootstrap.
Are there any issues with merging?
Would it be wise to wait a few days in case @catap wants to advise on the patch?
[main.log](https://github.com/user-attachments/files/23813853/main.log)


barracuda156:
@catap Kirill, are we good to merge this? Low-risk patch, since we do not declare support for 10.4, but still better to review this.

catap:
@ForestExpertise it makes sense, yes.

ForestExpertise:
> @ForestExpertise it makes sense, yes.

Thank you for your work on it, and for confirming it makes sense.
@barracuda156 Are we safe to merge?

barracuda156:
@ForestExpertise Yeah, this can be merged, just please move the patch after xar one (i.e. after general patches) and perhaps remove 0014 number from its name, it is out of context.

ForestExpertise:
> @ForestExpertise Yeah, this can be merged, just please move the patch after xar one (i.e. after general patches) and perhaps remove 0014 number from its name, it is out of context.

Should be fixed now!

barracuda156 (approved):

barracuda156:
Thank you!

PR #29, opened 2025-11-29
libxmlb: Fix build on Tiger

ForestExpertise:
I ran into a similar error to what was described in this ticket: https://trac.macports.org/ticket/63469 And one of the solutions suggested in that thread works easily.
The same solution also fixes libadwaita. Some different ones are needed for appstream, like adding the legacysupport portgroup. After my fixes to all three, Komikku functions, but is slow on a single processor G4.
Since libxmlb is the only port currently in powerpc ports, it is the only fix I am submitting for now.


barracuda156 (approved):
I am not testing this, but it is safe to merge.

PR #33, opened 2025-12-03
mldonkey: fix build on Tiger

ForestExpertise:
The 'const char **' error is common on Tiger.  An example with the solution I borrowed can be found here: https://trac.macports.org/ticket/71658 I can present the failed log if you think that might lead to a better solution, but mldonkey is working fine for me. Web GUI launches and is navigable. I don't know enough about these types of programs to know if all functionality is there.


barracuda156 (approved):

barracuda156:
Cool, thank you.

Does `any-dl` also build fine?

PR #34, opened 2025-12-08
ocaml-dune: add Tiger fallback

ForestExpertise:
As suggested in the other PR, I am adding a fallback from the commit right before #include libproc.h gets added.

### DIFF
--- a/ocaml/ocaml-dune/Portfile
+++ b/ocaml/ocaml-dune/Portfile
@@ -3,6 +3,8 @@
 PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           ocaml 1.1
+PortGroup           legacysupport 1.1
+legacysupport.newest_darwin_requires_legacy 8
 
 name                ocaml-dune
 github.setup        ocaml dune 3.20.2
@@ -59,16 +61,13 @@ patchfiles-append       patch-fix-3.20.diff
 
 # add fallback for Tiger that does not need libproc.h
 platform darwin 8 {
-    PortGroup           legacysupport 1.1
-
     name                ocaml-dune
     github.setup        ocaml dune 0d98de1cd607e99ae099e5c23482443d22570fbb
-    revision            1
 
     checksums           rmd160  7c086368485084f4a2d1521b7359f939306bca38 \
                         sha256  a4a6a4bbf422b4ee1baef3182576ab3eb64deded49a858ae83656916b0faa57c \
                         size    1777067
-github.tarball_from archive
+    github.tarball_from archive
 
     depends_build-append port:gmake
     build.cmd ${prefix}/bin/gmake


ocaml/ocaml-dune/Portfile
```diff
@@ -57,6 +57,26 @@ patchfiles-append       patch-copyfile_stubs.c-define-COPYFILE_CLONE.diff
 # https://github.com/ocaml/dune/issues/12576
 patchfiles-append       patch-fix-3.20.diff
 
+# add fallback for Tiger that does not need libproc.h
+platform darwin 8 {
+    PortGroup           legacysupport 1.1
+
+    name                ocaml-dune
+    github.setup        ocaml dune 0d98de1cd607e99ae099e5c23482443d22570fbb
+    revision            1
```
barracuda156: Please use some release version here, not a commit. (Unless the last release is broken or dated by years.)
If commit has to be used for some reason, it should still have a sensible version set.

ocaml/ocaml-dune/Portfile
```diff
@@ -57,6 +57,26 @@ patchfiles-append       patch-copyfile_stubs.c-define-COPYFILE_CLONE.diff
...
+
+    checksums           rmd160  7c086368485084f4a2d1521b7359f939306bca38 \
+                        sha256  a4a6a4bbf422b4ee1baef3182576ab3eb64deded49a858ae83656916b0faa57c \
+                        size    1777067
+github.tarball_from archive
```
barracuda156:
```
    github.tarball_from archive
```
(Alignment to 4 spaces.)

ocaml/ocaml-dune/Portfile
ForestExpertise: > Please use some release version here, not a commit. (Unless the last release is broken or dated by years.) If commit has to be used for some reason, it should still have a sensible version set.

Unfortunately, some ocaml-dune dependents check for the version of dune used, and error if it is not high enough. This version is superior to 3.3 for that reason alone, because it is checked as 3.4, which is needed for some of the dependencies. In addition, this commit is only days before the release of 3.4, so it is likely fairly stable. I have encountered no issues with it.
If I were more skilled, the ideal thing to do would be to patch ocaml-dune 3.15 (the minimum version required for one dependency I encountered) or later to use what they used before libproc.h, but that is beyond my skillset for now.
I did get rid of the revision number, which was an artifact of my previous attempts. I believe I have made all other requested changes.

barracuda156: > In addition, this commit is only days before the release of 3.4, so it is likely fairly stable.

Does it work to revert the breaking commit from 3.4 release then? You can check out to the release commit and run `git revert ${breaking_commit_hash}`. If that fails, see if it can be easily done manually.

barracuda156:
@ForestExpertise Could you please confirm if the update works for you?

PR #38, opened 2025-12-14
abiword-x11: fix Tiger build by building without boost

ForestExpertise:
If there is a way to not require boost at all, that would be even better. Failing that, I have tested this and it build.


barracuda156:
Hope this works

PR #50, opened 2026-01-02
FBReader: new port in office

ForestExpertise:
FBReader is a nifty epub reader with support for table of contents (unlike epy or mupdf). This was the last version to have qt4 gui support for Mac. Building with cocoa for 10.5+ should be possible if desired, but is not something easily doable on Tiger. Theoretically, gtk support may also be possible (it exists for Linux in the source), but I don't see the point. Tested and functional, though it's possible an environmental variable has to be set (I think I fixed that bug, but am unsure). Needs ccache and unibreak, which is why I fixed those earlier.

### DIFF
--- a/office/FBReader/Portfile
+++ b/office/FBReader/Portfile
@@ -4,11 +4,11 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           legacysupport 1.1
 PortGroup           qt4 1.0
-PortGroup           qmake 1.0
 
 name                FBReader
 github.setup        geometer FBReader 0.99.5
-github.tarball_from tarball
+revision            2
+github.tarball_from archive
 categories          office
 description         A lightweight ebook reader.
 long_description    ${description}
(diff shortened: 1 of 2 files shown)


office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+depends_build-append \
+                    port:ccache \
+                    port:gmake
+
+depends_lib         port:curl \
+                    port:expat \
+                    port:fribidi \
+                    port:libunibreak \
+                    port:libzip \
+                    port:qt4-mac \
```
barracuda156: No need, PG already sets that. (Unless you mean that the port is broken with `qt-mac-devel` and must use exactly `qt4-mac`.

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           legacysupport 1.1
+PortGroup           qt4 1.0
+PortGroup           qmake 1.0
+
+name                FBReader
+github.setup        geometer FBReader 0.99.5
+github.tarball_from tarball
```
barracuda156: Please use `archive` or `release`, since `tarball` is deprecated.

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           legacysupport 1.1
+PortGroup           qt4 1.0
+PortGroup           qmake 1.0
```
barracuda156: Either one of the other. `qmake` is used when the build is done with `qmake`. `qt4` is used when the build is not done with `qmake`.
It seems you need `qt4` here.

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+
+patchfiles-append   target-mk.patch
+
+platform darwin 8 {
+patchfiles-append   iconv.patch
+}
+
+configure {}
+
+build.env-append prefix=${prefix} \
```
barracuda156: `${prefix}/libexec/qt4` should probably be `${qt_dir}` (it may be installed elsewhere), though admittedly Qt4 handling is somewhat confusing. Check qt4 and qt4-mac PGs for details.

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+                 FBREADERQTLIB_FLAGS=${prefix}/libexec/qt4/lib \
+                 FBREADERQTFRAME_FLAGS=${prefix}/libexec/qt4/Library/Frameworks/QtCore.framework/Versions/4 \
+                 FBREADERIQT_FLAGS=${prefix}/libexec/qt4/include \
+                 FBREADERIQTCORE_FLAGS=${prefix}/libexec/qt4/include/QtCore \
+                 FBREADERIQTGUI_FLAGS=${prefix}/libexec/qt4/include/QtGui \
+                 FBREADERIQTNETWORK_FLAGS=${prefix}/libexec/qt4/include/QtNetwork \
+                 FBREADERQTINCLUDE_FLAGS=${prefix}/libexec/qt4/include \
+                 FBREADERFFREAMEWORKS_FLAGS=${prefix}/libexec/qt4/Library/Frameworks \
+                 FBREADERFINCLUDE_FLAGS=${prefix}/libexec/qt4/include \
+                 FBREADERFLIB_FLAGS=${prefix}/libexec/qt4/lib
```
barracuda156: Doesn’t it find specific libs, once the Qt root directory is pointed at?

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+system "install_name_tool -change ${prefix}/lib/libgcc/libstdc++.6.dylib \
+    @executable_path/../Frameworks/libstdc++.6.dylib ${fbreader}"
+system "install_name_tool -change ${prefix}/lib/libgcc/libgcc_s.1.1.dylib \
+    @executable_path/../Frameworks/libgcc_s.1.1.dylib ${fbreader}"
+system "cd ../../../../../../../../../../.. \
+sudo ${prefix}/libexec/qt4/bin/macdeployqt ${prefix}/var/macports/build/_Users_powerpc-ports_office_FBReader/FBReader/work/destroot/Applications/Macports/FBReader.app"
+}
+
+post-activate {
+system "sudo install_name_tool -change @executable_path/../Frameworks/libgcc_s.1.1.dylib ${prefix}/lib/libgcc/libgcc_s.1.1.dylib /Applications/Macports/FBReader.app/Contents/MacOS/FBReader"
```
barracuda156: And then reverted?

office/FBReader/files/target-mk.patch
```diff
@@ -0,0 +1,87 @@
...
++FBREADERIQT_FLAGS=$(prefix)/libexec/qt4/include \
++FBREADERIQTCORE_FLAGS=$(prefix)/libexec/qt4/include/QtCore \
++FBREADERIQTGUI_FLAGS=$(prefix)/libexec/qt4/include/QtGui \
++FBREADERIQTNETWORK_FLAGS=$(prefix)/libexec/qt4/include/QtNetwork \
++FBREADERQTINCLUDE_FLAGS=$(prefix)/libexec/qt4/include \
++FBREADERFFREAMEWORKS_FLAGS=$(prefix)/libexec/qt4/Library/Frameworks \
++FBREADERFINCLUDE_FLAGS=$(prefix)/libexec/qt4/include \
++FBREADERFLIB_FLAGS=$(prefix)/libexec/qt4/lib
+-ARCH_FLAGS = -arch x86_64 -mmacosx-version-min=10.5
++ARCH_FLAGS = -arch ppc -mmacosx-version-min=10.4
```
barracuda156: This should be picked from flags passed by the base. The only thing needed is to ensure environment flags are actually used. If that is problematic, then they are to be patched in, but not hardcoded.

office/FBReader/Portfile
```diff
@@ -0,0 +1,86 @@
...
+foreach dep [exec otool -L ${fbreader} | grep libgcc] {
+    if [string match -nocase "*/libgcc*.dylib" {dep}] {
+        ui_info "Copy libgcc files over from ${dep} to {worksrcpath}/Contents/Frameworks"
+        file copy ${dep} ${workpath}/destroot/Applications/Contents/Frameworks
+    }
+}
+system "install_name_tool -change ${prefix}/lib/libgcc/libstdc++.6.dylib \
+    @executable_path/../Frameworks/libstdc++.6.dylib ${fbreader}"
+system "install_name_tool -change ${prefix}/lib/libgcc/libgcc_s.1.1.dylib \
+    @executable_path/../Frameworks/libgcc_s.1.1.dylib ${fbreader}"
```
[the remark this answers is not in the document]
ForestExpertise: There is a bug on macdeployqt on older versions where it can't seem to find some gcc related dylibs and thus won't launch. https://github.com/macports/macports-ports/commit/1adb4cc7c2b3183f3d2374196deab8a2542f2380 addresses the same problem, and I adapted it to the needs of this program.

office/FBReader/Portfile
ForestExpertise: This is actually done again each activation, because at least for this program, if you deactivate and then activate again, it forgets where the dylibs are. Hence the post-activate step.

office/FBReader/files/target-mk.patch
ForestExpertise: The base fails to pass, so I put it in an explicit environmental variable.

barracuda156:
@ForestExpertise Allow me to test this locally and sort out some things. It is understandably difficult, and I can’t give meaningful suggestions on some details without seeing what the build actually does.

barracuda156:
@ForestExpertise Could you try this? https://github.com/macos-powerpc/powerpc-ports/commit/9e65ca87c443419c86b20cd6aa3433067f85bae3

Looks like it works on 10.6.8. (On Catalina I get a crash on trying to access online libraries. No idea why and won’t debug now.)

ForestExpertise:
> @ForestExpertise Could you try this? [8e0e544](https://github.com/macos-powerpc/powerpc-ports/commit/8e0e54448071fcc55f617c9433da20bf4fc2615b)
> 
> Looks like it works on 10.6.8. (On Catalina I get a crash on trying to access online libraries. No idea why and won’t debug now.)

Unfortunately it does not work for me, I attached a screenshot because I am in a rush and about to travel
IMG_20260130_210956886
I can try to sort when I get back, the error looks familiar, I feel like it may be something Tiger specific I have seen and fixed before

barracuda156:
@ForestExpertise I decided to commit the version which I had first, there is little point to be conservative with end-user non-essential ports with no dependents; if we need some extra fixes for 10.4 or just improvements, please submit a new PR. If you want to maintain the port, I will replace `nomaintainer` with your GH handle.

barracuda156:
I am not completely sure the problem is specific to Tiger. Qt4 portgroups are complicated, and there are options where to install. Possibly just defaults work differently.

Though variables should work regardless (hardcoded paths may not).

PR #51, opened 2026-01-02
PortGroup boost-1.0.tcl: Move to 1.76 for Tiger and Leopard

ForestExpertise:
Since boost176 (and boost181, I did test it) now works on Tiger, and boost171 still doesn't, could we move to boost176 as the default? If Leopard can build boost176, perhaps all powerpc-ports could move to boost portgroup being a stub for boost176 instead of boost171.

### DIFF
--- a/_resources/port1.0/group/boost-1.0.tcl
+++ b/_resources/port1.0/group/boost-1.0.tcl
@@ -28,10 +28,7 @@ default boost_cache_env_vars      ""
 
 proc boost::default_version {} {
     global os.platform os.major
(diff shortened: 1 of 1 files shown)


barracuda156:
I am surprised 1.71 was the default for Leopard LOL

Let’s just drop 1.71 completely from there. This will be a breaking change, but carrying 1.71 indefinitely is not great either.
I will handle revbumping for < 10.6 for our ports which use default boost; besides those, users will have to force-upgrade any other port which they have built against 1.71.

PR #52, opened 2026-01-03
miniupnpc: fix missing include

Matias314:
Miniupnpc release has a small defect: a missing include in upnpcommands.h which broke the build of transmission-qt and possibly other things. 

This has been rectified upstream but there hasn't been a new release in 8 months and it's unclear when there finally will be so I thought it would be a good idea to include this patch until at least the next release. Please let me know your thoughts.


barracuda156 (approved):
Thank you

PR #61, opened 2026-01-06
elinks-devel: fix build on Tiger

ForestExpertise:
execinfo.h does not exist on Tiger. Disabling backtrace keeps it from being included and causing a build failure.

### DIFF
--- a/www/elinks-devel/Portfile
+++ b/www/elinks-devel/Portfile
@@ -119,8 +119,9 @@ compiler.cxx_standard       2017
 
 # avoid execinfo.h
 platform darwin 8 {
-    configure.args-delete   -Dbacktrace=true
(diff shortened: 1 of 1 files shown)


www/elinks-devel/Portfile
```diff
@@ -117,6 +117,12 @@ configure.args-append \
 
 compiler.cxx_standard       2017
 
+# avoid execinfo.h
+platform darwin 8 {
+    configure.args-delete   -Dbacktrace=true
+    configure.args-append   -Dbacktrace=false
```
barracuda156: Could you use `replace`, and we will merge.

```
    configure.args-replace \
                    -Dbacktrace=true \
                    -Dbacktrace=false
```

PR #77, opened 2026-02-18
libheif: remove optional vvdec dependency on Tiger

ForestExpertise:
Vvdec does not build on Tiger due to usage of posix_memalign Since this is the only port that depends on it, and vvdec is optional for libheif, it is simplest to remove it.


barracuda156:
@ForestExpertise I do not oppose dropping it, but could you confirm it still fails? This might have fixed it: https://github.com/macos-powerpc/powerpc-ports/commit/e7f0d25cc3a05f1093425edc32b8d9af64139012

PR #101, opened 2026-05-02
Fastfetch: update 0016-Tiger-specific-adjustments-to-CMakeLists.patch

Matias314:
The CMakeLists.txt was changed a bit causing the old patch to no longer apply. This fixes the build on 10.5


barracuda156 (changes requested):
Please change commit message. It should start from port name, i.e. `fastfetch: `. We want commit history to be readable.

Matias314:
Done

barracuda156 (approved):

PR #103, opened 2026-05-07
pulseaudio: fix sample format on big endian, revbump

Matias314:
I  don't know if this is the right way to do it, but it works. Have working audio on 10.5 now instead of just static.


barracuda156 (approved):
Let’s try it.

PR #111, opened 2026-05-14
QMPlay2: Fix Tiger build

ForestExpertise:
Relevant to all OS - I restored variants, which is important for low end G4s and G5s.
QT4 from powerpc-ports installs the plugins directory to a different place on 10.4 than on 10.5 or higher, but that is easy to patch in relevant ports with reinplace. The screensaver code seems too much effort to fix properly, so I converted into stubs. I also fixed the ambiguous OpenGL call, but I am still not seeing better performance from OpenGL2 than software rendering (480p, Youtube, h.264 preferred). Could be a Tiger issue, or more likely a 32MB of VRAM issue. I will test on a PowerBook with 128 MB of VRAM later and see if that improves OpenGL2 enough to make it worth using. On that same hardware I have a Leopard partition so I can see if Tiger is the issue instead of the hardware.

### DIFF
--- a/multimedia/QMPlay2/Portfile
+++ b/multimedia/QMPlay2/Portfile
@@ -78,12 +78,13 @@ if {${os.platform} eq "darwin" && ${os.major} < 15} {
     # configure.cppflags-append \
                     -DUSE_QRAWFONT
 
-    # FIXME: specifically on Tiger this fails:
-    # OpenGL2Common.cpp:258:44: error: call of overloaded 'setUniformValue(const char*, int)' is ambiguous
-    # Disable for now, GL is optional anyway.
     platform darwin 8 {
-        configure.args-replace \
-                    -DUSE_OPENGL2=ON -DUSE_OPENGL2=OFF
+        patchfiles-append \
+                    tiger-fixes.patch
+        post-patch {
+            reinplace "s|@qt_tiger_plugins_dir@|${prefix}/share/qt4/plugins|" ${worksrcpath}/src/gui/CMakeLists.txt
+        }
+        configure.cxxflags-append -fpermissive
     }
 
     post-destroot {
@@ -290,4 +291,41 @@ variant pulse description "Use Pulseaudio" {
                     port:pulseaudio
     configure.args-replace \
                     -DUSE_PULSEAUDIO=OFF -DUSE_PULSEAUDIO=ON
+}
+
+variant lto description "Enable LTO" {
+    # Does not seem to do anything on its own.
+    configure.args-replace \
+                    -DUSE_LINK_TIME_OPTIMIZATION=OFF \
+                    -DUSE_LINK_TIME_OPTIMIZATION=ON
+    # So just pass flags explicitly:
+    configure.cxxflags-append \
+                    -flto
+    configure.ldflags-append \
+                    -flto
+}
+
+variant G4 conflicts G5 description "Optimize for G4" {
+    configure.cxxflags-append \
+                    -mcpu=G4 -mtune=G4
+}
+
+variant G5 conflicts G4 description "Optimize for G5" {
+    configure.cxxflags-append \
+                    -mcpu=970 -mtune=970
+
+    # https://github.com/iains/gcc-14-branch/issues/24
+    if {${configure.build_arch} ne "ppc64"} {
+        configure.cxxflags-append \
+                    -mno-powerpc64
+    }
+}
+
+if {${os.platform} ne "darwin" || ${os.major} > 9} {
+    default_variants-append +lto
+}
+
+# ppc64 can only be built on a G5, so default to that.
+if {${os.platform} eq "darwin" && ${configure.build_arch} eq "ppc64"} {
+    default_variants-append +G5
 }
--- a/multimedia/QMPlay2/Portfile
+++ b/multimedia/QMPlay2/Portfile
@@ -81,9 +81,6 @@ if {${os.platform} eq "darwin" && ${os.major} < 15} {
     platform darwin 8 {
         patchfiles-append \
                     tiger-fixes.patch
-        post-patch {
-            reinplace "s|@qt_tiger_plugins_dir@|${prefix}/share/qt4/plugins|" ${worksrcpath}/src/gui/CMakeLists.txt
-        }
         configure.cxxflags-append -fpermissive
     }
 
@@ -291,41 +288,4 @@ variant pulse description "Use Pulseaudio" {
                     port:pulseaudio
     configure.args-replace \
                     -DUSE_PULSEAUDIO=OFF -DUSE_PULSEAUDIO=ON
-}
-
-variant lto description "Enable LTO" {
-    # Does not seem to do anything on its own.
-    configure.args-replace \
-                    -DUSE_LINK_TIME_OPTIMIZATION=OFF \
-                    -DUSE_LINK_TIME_OPTIMIZATION=ON
-    # So just pass flags explicitly:
-    configure.cxxflags-append \
-                    -flto
-    configure.ldflags-append \
-                    -flto
-}
-
-variant G4 conflicts G5 description "Optimize for G4" {
-    configure.cxxflags-append \
-                    -mcpu=G4 -mtune=G4
-}
-
-variant G5 conflicts G4 description "Optimize for G5" {
-    configure.cxxflags-append \
-                    -mcpu=970 -mtune=970
-
-    # https://github.com/iains/gcc-14-branch/issues/24
-    if {${configure.build_arch} ne "ppc64"} {
-        configure.cxxflags-append \
-                    -mno-powerpc64
-    }
-}
-
-if {${os.platform} ne "darwin" || ${os.major} > 9} {
-    default_variants-append +lto
-}
-
-# ppc64 can only be built on a G5, so default to that.
-if {${os.platform} eq "darwin" && ${configure.build_arch} eq "ppc64"} {
-    default_variants-append +G5
 }
--- a/multimedia/QMPlay2/files/tiger-fixes.patch
+++ b/multimedia/QMPlay2/files/tiger-fixes.patch
@@ -42,18 +42,6 @@
          shaderProgramVideo->setUniformValue("uTextureSize", m_textureSize);
  
          doReset = !resetDone;
---- a/src/gui/CMakeLists.txt
-+++ b/src/gui/CMakeLists.txt
-@@ -231,8 +231,8 @@
- elseif(APPLE)
-     install(TARGETS ${PROJECT_NAME} BUNDLE DESTINATION ${CMAKE_INSTALL_PREFIX})
- 
-     set(QT_LIBS_DIR "@qt_libs_dir@")
--    set(QT_PLUGINS_DIR "@qt_plugins_dir@")
-+    set(QT_PLUGINS_DIR "@qt_tiger_plugins_dir@")
- if(NOT USE_QT4)
-     install(FILES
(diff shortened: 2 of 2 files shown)


barracuda156:
@ForestExpertise Thank you. Looks good generally, but I wanna investigate a bit re Qt4 installation layout. This problem keeps popping up, and we rather avoid making changes under potentially inaccurate assumptions. Could you clarify the following:

1. Have you got `qt4-mac` and `qt4-mac-devel` both installed or only one of these? (They cannot be active simultaneously, but can be installed.)
2. Just to confirm, here `qt4-mac` is used?
3. Have you installed `qt4-mac` from my port after installing PPCPorts or was it installed before that? If you use a custom base (which is probably the case, given Tiger), then which exactly version? What I am specifically interested in is a) if your base uses portgroups like PPCPorts or like upstream MacPorts; b) if identical setup was used when installing `qt4-mac` and `QMPlay2`. (I do not mean they must be identical, I just wanna figure out, why we get such a mess.)
4. Could you show exact paths on your system for relevant directories (`port contents qt4-mac`, then find where plugins are).

It does not seem plausible to me that we get different paths based on OS version (why would it be the case?).
How upstream MacPorts installs can be seen from https://ports.macports.org/port/qt4-mac/details (no cases of 10.4/10.5, of course). What I suspect is that either ports do something a bit differently or port groups.

ForestExpertise:
1. Only qt4-mac is installed, qt4-mac-devel does not currently build on Tiger, and with qt4-mac working, is not a high priority for me.
2. qt4-mac is used here, from PowerPC ports. The recent update fixed it for Tiger, thank you again.
3. The base is from the tiger_upd branch of https://github.com/macos-powerpc/ppcports-base it builds without issues from source, but is probably closest to 2.11.6 as far as I can tell.
4. I will hopefully get a chance to run port contents qt4-mac in about 5 hours and update you

It is very plausible that we get different paths based on OS version because 10.4 does not support rpath, while 10.5 and later do. Thanks to the improvements to ld64-97, Tiger can avoid erroring on rpath, but does not implement it properly, which can lead to files in weird places.
Edit: On second thought, I suppose qt4-mac could be installing differently than qt4-mac-devel, which would make this not an OS issue, but it seems unlikely. My guess is some sort of Tiger weirdness, maybe rpath related.

barracuda156:
Well, rpath support as such has no influence on where files are installed, installation is controlled by install target in makefiles (or their equivalent) and portfile code (plus portgroups if used). While it is conceivable that some specific OS version _should_ install files differently, that won’t happen automatically. If `qt4-mac`, Qt4 portgroups and install targets in Qt4 code do not have special cases for 10.4, paths are expected to match (what may happen is that some extra files are present or some missing, but whatever coincides should also have identical paths).

`qt4-mac` and `qt4-mac-devel` do install differently on my end and likely generally, which is a pain, but I do not want to touch it (I mean, just look at the amount of code affecting that). What I am not sure is whether `qt4-mac` itself installs identically from PPCPorts vs MacPorts for every given configuration. If that differs, we get an extra problem: using the upstream base instead of PPCPorts base changes which portgroups are used, and ports themselves diverge, so if you install `qt4-mac` with standard MacPorts and then upgrade to PPCPorts, installation layout may differ from what is assumed.

ForestExpertise:
As you suggested was a possibility, the issue is not Tiger, but qt4-mac. Port contents reveals the plugins directory to be opt/local/share/qt4/plugins on both my Tiger and Leopard partitions. I suspect you can replicate this issue by trying to use qt4-mac to build Qmplay2 instead of qt4-mac-devel, and that you can replicate the same plugins directory location.

barracuda156:
Ok, thanks, I will check.

P. S. I doubt this was added completely blindly: https://github.com/macos-powerpc/powerpc-ports/blob/4810b75781e87b8551da038b6a34f3cc60bab6e4/multimedia/QMPlay2/Portfile#L31-L38
I also certainly did build the port locally:
https://macos-powerpc.org/packages/QMPlay2/QMPlay2-19.09.03_0.darwin_10.ppc.tbz2 (10.6)
https://macos-powerpc.org/packages_ppc64/QMPlay2/QMPlay2-19.09.03_0.darwin_9.ppc64.tbz2 (10.5)
If you open an archive, it has portfile which was used.
So I do not know what is going on.

barracuda156:
@ForestExpertise Ok, so here is what I have on macOS 10.15:

`qt4-mac` @4.8.7_16:
```
  /opt/local/libexec/qt4/share/plugins/accessible/libqtaccessiblecompatwidgets.dylib
  /opt/local/libexec/qt4/share/plugins/accessible/libqtaccessiblewidgets.dylib
  /opt/local/libexec/qt4/share/plugins/bearer/libqcorewlanbearer.dylib
  /opt/local/libexec/qt4/share/plugins/bearer/libqgenericbearer.dylib
  /opt/local/libexec/qt4/share/plugins/codecs/libqcncodecs.dylib
  /opt/local/libexec/qt4/share/plugins/codecs/libqjpcodecs.dylib
  /opt/local/libexec/qt4/share/plugins/codecs/libqkrcodecs.dylib
  /opt/local/libexec/qt4/share/plugins/codecs/libqtwcodecs.dylib
  /opt/local/libexec/qt4/share/plugins/designer/libqdeclarativeview.dylib
  /opt/local/libexec/qt4/share/plugins/designer/libqt3supportwidgets.dylib
  /opt/local/libexec/qt4/share/plugins/designer/libqwebview.dylib
  /opt/local/libexec/qt4/share/plugins/graphicssystems/libqglgraphicssystem.dylib
  /opt/local/libexec/qt4/share/plugins/graphicssystems/libqtracegraphicssystem.dylib
  /opt/local/libexec/qt4/share/plugins/iconengines/libqsvgicon.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqgif.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqico.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqjpeg.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqmng.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqsvg.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqtga.dylib
  /opt/local/libexec/qt4/share/plugins/imageformats/libqtiff.dylib
  /opt/local/libexec/qt4/share/plugins/qmltooling/libqmldbg_inspector.dylib
  /opt/local/libexec/qt4/share/plugins/qmltooling/libqmldbg_tcp.dylib
  /opt/local/libexec/qt4/share/plugins/script/libqtscriptdbus.dylib
```

`qt4-mac-devel` @4.8.7_8
```
  /opt/local/share/qt4/plugins/accessible/libqtaccessiblecompatwidgets.dylib
  /opt/local/share/qt4/plugins/accessible/libqtaccessiblewidgets.dylib
  /opt/local/share/qt4/plugins/bearer/libqcorewlanbearer.dylib
  /opt/local/share/qt4/plugins/bearer/libqgenericbearer.dylib
  /opt/local/share/qt4/plugins/codecs/libqcncodecs.dylib
  /opt/local/share/qt4/plugins/codecs/libqjpcodecs.dylib
  /opt/local/share/qt4/plugins/codecs/libqkrcodecs.dylib
  /opt/local/share/qt4/plugins/codecs/libqtwcodecs.dylib
  /opt/local/share/qt4/plugins/designer/libqdeclarativeview.dylib
  /opt/local/share/qt4/plugins/designer/libqt3supportwidgets.dylib
  /opt/local/share/qt4/plugins/graphicssystems/libqglgraphicssystem.dylib
  /opt/local/share/qt4/plugins/graphicssystems/libqtracegraphicssystem.dylib
  /opt/local/share/qt4/plugins/iconengines/libqsvgicon.dylib
  /opt/local/share/qt4/plugins/imageformats/libqgif.dylib
  /opt/local/share/qt4/plugins/imageformats/libqico.dylib
  /opt/local/share/qt4/plugins/imageformats/libqjpeg.dylib
  /opt/local/share/qt4/plugins/imageformats/libqmng.dylib
  /opt/local/share/qt4/plugins/imageformats/libqsvg.dylib
  /opt/local/share/qt4/plugins/imageformats/libqtga.dylib
  /opt/local/share/qt4/plugins/imageformats/libqtiff.dylib
  /opt/local/share/qt4/plugins/qmltooling/libqmldbg_inspector.dylib
  /opt/local/share/qt4/plugins/qmltooling/libqmldbg_tcp.dylib
  /opt/local/share/qt4/plugins/script/libqtscriptdbus.dylib
  /opt/local/share/qt4/plugins/sqldrivers/libqsqlite.dylib
```

I am not sure what to do. I hoped that we just got two cases (which is already bad), but turns out it is not guaranteed that even `qt4-mac` installs identically.
Possible kinda non-invasive options seem to be:
1) install a symlink with qt4-mac* ports; downside: everyone needs to rebuild them (or use a local hack);
2) add a logic to ports that use Qt4 plugins (hopefully not many of these) to check for presence of needed directory and adjust paths accordingly.

ForestExpertise:
Thank you for investigating. I am glad to hear that it wasn't Tiger or myself doing something wrong, but that this seems to be a qt4-mac bug.
I am in favor of rebuilding and sym linking, personally, it seems easier than coming up with logic to find it installed directories. But whichever option you choose will be an improvement.

barracuda156:
@ForestExpertise If you prefer, we can first deal with other changes and postpone Qt4 plugins stuff a bit. Arguably there are no too many users on Tiger, and if someone complains, we can point to your solution. I need to make sure that whatever changes are made, work across different systems, so that we avoid breaking something that works for someone else, while fixing this case.

> but that this seems to be a qt4-mac bug.

I am not sure at the point that it is a bug, it could be by design for a credible reason. I am not the author of any Qt4 portgroups, whether in upstream MacPorts or here (I borrowed them from macstrop with cosmetic changes), so I don’t really know their internals.

barracuda156:
P. S. Be sure that this situation with Qt4 is not something I enjoy myself lol, so I will try to figure out how to make it work.

barracuda156:
@RJVB This issue keeps haunting us. Would a symlink be an appropriate solution? I don’t really get how the portgroup decides where to install stuff, and while forcing identical paths for both ports is not hard, that may lead to a havoc with dependents, and I do not want to rebuild dozens of huge ports.

ForestExpertise:
> @ForestExpertise If you prefer, we can first deal with other changes and postpone Qt4 plugins stuff a bit. Arguably there are no too many users on Tiger, and if someone complains, we can point to your solution. I need to make sure that whatever changes are made, work across different systems, so that we avoid breaking something that works for someone else, while fixing this case.
> 
> > but that this seems to be a qt4-mac bug.
> 
> I am not sure at the point that it is a bug, it could be by design for a credible reason. I am not the author of any Qt4 portgroups, whether in upstream MacPorts or here (I borrowed them from macstrop with cosmetic changes), so I don’t really know their internals.

None of this is urgent, I am happy to drop the part related to differing QT4 install paths and just have the rest merged. Let me know if I should go ahead with that.

barracuda156:
If you don’t mind, it would be nice to have two commits now: 1) Tiger fixes (without Qt4 plugins part); 2) improvements (added variants). So that we don’t get variants popping up out of nowhere without an indication in commit history.

Qt4 part will be dealt with separately.

ForestExpertise:
This pull request should work for the commit to fix Tiger build, I will re add variants in another pull request after this gets merged. I have removed QT4 stuff as hopefully that can be fixed elsewhere, maybe with your excellent symlink plan

barracuda156 (approved):

PR #114, opened 2026-05-17
ffmpeg, ffmpeg4, ffmpeg5: add missing kaudioformats on 10.5, fixes build

Matias314:
ps: why do we have both `ffmpeg` and `ffmpeg4`?

### DIFF
--- a/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
+++ b/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
@@ -30,4 +30,5 @@
 +#define kAudioFormatiLBC 'ilbc'
 +#endif
 
- typedef struct ATDecodeConte
+ typedef struct ATDecodeContext {
+     AVClass *av_class;
--- a/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
+++ b/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
@@ -1,10 +1,16 @@
 --- libavcodec/audiotoolboxdec.c.orig
 +++ libavcodec/audiotoolboxdec.c
-@@ -36,6 +36,11 @@
+@@ -31,11 +31,17 @@
+ #include "libavutil/avassert.h"
+ #include "libavutil/opt.h"
+ #include "libavutil/log.h"
++#include <AvailabilityMacros.h>
+
+ #if __MAC_OS_X_VERSION_MIN_REQUIRED < 101100
  #define kAudioFormatEnhancedAC3 'ec-3'
  #endif
 
-+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060
++#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
 +#define kAudioFormatMicrosoftGSM 0x6D730031
 +#define kAudioFormatiLBC 'ilbc'
 +#endif
@@ -14,14 +20,14 @@
 
 --- libavcodec/audiotoolboxenc.c.orig
 +++ libavcodec/audiotoolboxenc.c
-@@ -35,6 +35,10 @@
+@@ -34,6 +34,11 @@
+ #include "libavutil/avassert.h"
  #include "libavutil/opt.h"
  #include "libavutil/log.h"
-
-+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060
++#include <AvailabilityMacros.h>
++
++#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
 +#define kAudioFormatiLBC 'ilbc'
 +#endif
-+
- typedef struct ATDecodeContext {
-     AVClass *av_class;
-     int mode;
+
+ typedef struct ATDecodeConte
--- a/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
+++ b/multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
@@ -1,0 +1,27 @@
+--- libavcodec/audiotoolboxdec.c.orig
++++ libavcodec/audiotoolboxdec.c
+@@ -36,6 +36,11 @@
+ #define kAudioFormatEnhancedAC3 'ec-3'
+ #endif
+
++#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060
++#define kAudioFormatMicrosoftGSM 0x6D730031
++#define kAudioFormatiLBC 'ilbc'
++#endif
++
+ typedef struct ATDecodeContext {
+     AVClass *av_class;
+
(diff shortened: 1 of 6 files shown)


barracuda156:
> ps: why do we have both `ffmpeg` and `ffmpeg4`?

You mean “we” abstractly or in PPCPorts? I suspect MacPorts added `ffmpeg4` to switch unmaintained software to it (to avoid fixing every build for new APIs), having in view a general move to `ffmpeg8`. And while I am not particularly fascinated with the idea, it is far easier just to match MacPorts than to pick every port using ffmpeg and configure it differently.
(And _not_ having some ffmpeg version in PPCPorts would mean that we cannot update any dependencies that require revbumps or control variants, and every such update done on MacPorts end gonna mess things for us.)

Matias314:
Ahh okay. I was just curious because they appear to be the same version, but the portfiles contain quite a few differences.

By the way I am curious of your opinion for this fix. Because while [upstream](https://code.ffmpeg.org/FFmpeg/FFmpeg/src/branch/release/4.4/libavcodec/audiotoolboxdec.c#L35) does something similar to what I did, I see that you handled it differently by guarding the cases in the switch statement: https://github.com/macos-powerpc/powerpc-ports/blob/3d07df75fb4f713162a90ad3b8bf1984e9dde5eb/multimedia/ffmpeg/files/patch-libavcodec-audiotoolboxenc.c.diff

barracuda156 (commented):
@Matias314 Do these files import `Availability.h` and not `AvailabilityMacros.h`? This form of macros `#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060` is unsafe, since it can expand to `#if 0 < 1060` (when `__MAC_OS_X_VERSION_MIN_REQUIRED` is undefined), and we get redefinitions where they should not be. Preferable practice is to use `AvailabilityMacros.h` and unprefixed macros. Unless a source already uses underscore-prefixed ones _and_ imports `Availability.h` specifically (directly or transitively).

barracuda156:
> Ahh okay. I was just curious because they appear to be the same version, but the portfiles contain quite a few differences.

I made some changes when adding `ffmpeg4`. There is no requirement to keep these identical anyway, switching from one to another requires revbumping and therefore rebuilds.

> By the way I am curious of your opinion for this fix. Because while [upstream](https://code.ffmpeg.org/FFmpeg/FFmpeg/src/branch/release/4.4/libavcodec/audiotoolboxdec.c#L35) does something similar to what I did, I see that you handled it differently by guarding the cases in the switch statement: https://github.com/macos-powerpc/powerpc-ports/blob/3d07df75fb4f713162a90ad3b8bf1984e9dde5eb/multimedia/ffmpeg/files/patch-libavcodec-audiotoolboxenc.c.diff

This is probably not my patch, and it follows upstream, just moving threshold from 10.6 to 10.7: https://github.com/FFmpeg/FFmpeg/blob/989e621bcd93c3c79dbfbe65710505259b8c73e9/libavcodec/audiotoolboxenc.c#L73-L76
Upstream seems to be wrong: https://developer.apple.com/documentation/coreaudiotypes/kaudioformatmpeg4aac_eld (well, Apple can be wrong too, but I guess the patch exists for a reason).

Matias314:
> @Matias314 Do these files import Availability.h and not AvailabilityMacros.h? This form of macros #if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060 is unsafe, since it can expand to #if 0 < 1060 (when __MAC_OS_X_VERSION_MIN_REQUIRED is undefined), and we get redefinitions where they should not be. Preferable practice is to use AvailabilityMacros.h and unprefixed macros. Unless a source already uses underscore-prefixed ones and imports Availability.h specifically (directly or transitively).

Wait I'm confused, because [here](https://github.com/FFmpeg/FFmpeg/blob/989e621bcd93c3c79dbfbe65710505259b8c73e9/libavcodec/audiotoolboxdec.c#L39-L75) both with and without underscores are being used. At a first glance I don't see either `AvailabilityMacros.h` or `Availability.h` anywhere in the code. Is this set dynamically?

barracuda156:
It looks like a mess, one file even includes both: https://github.com/FFmpeg/FFmpeg/blob/989e621bcd93c3c79dbfbe65710505259b8c73e9/libavcodec/videotoolbox.c#L42-L43
AFAIK it is safe to include `AvailabilityMacros.h` even if it may already be pulled in indirectly, so I suggest adding this header and using macros without underscores.

barracuda156:
@Matias314 I am happy to merge this once that technicality is addressed, we certainly want ffmpeg ports to build.

Besides, do we need the same for other ffmpeg versions?

Matias314:
Yes sorry, I will address this soon. Too little time, I need more than 24 hours in a day to fit everything lol.

I only noticed the problem on these specific versions, but I would not be suprised if other versions will also develop the issue. Will have to keep an eye on it. I can't build ffmpeg8 for a different reason, but I think that can be fixed too.

barracuda156:
@Matias314 If `opencolorio` fails to build, we can just disable it for < 10.6.

Matias314:
I pushed the changes, apologies for the delay.

multimedia/ffmpeg4/files/patch-leopard-kaudioformat.diff
```diff
@@ -14,14 +20,14 @@
...
++
++#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
 +#define kAudioFormatiLBC 'ilbc'
 +#endif
-+
- typedef struct ATDecodeContext {
-     AVClass *av_class;
-     int mode;
+
+ typedef struct ATDecodeConte
```
barracuda156: @Matias314 Looks like here (and the same place in other patches) ending part got accidentally deleted?

Matias314:
Wait I don't see it. The end result is the same?

barracuda156:
> Wait I don't see it. The end result is the same?

This looks incomplete (two lines instead of three after the modified one, and `ATDecodeConte` looks wrong):
```
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
+#define kAudioFormatiLBC 'ilbc'
+#endif

 typedef struct ATDecodeConte
```
Earlier version had:
```
+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1060
+#define kAudioFormatiLBC 'ilbc'
+#endif
+
 typedef struct ATDecodeContext {
     AVClass *av_class;
     int mode;
```

Correct is probably this (based on updated version):
```
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
+#define kAudioFormatiLBC 'ilbc'
+#endif

 typedef struct ATDecodeContext {
     AVClass *av_class;
```

Matias314:
Thank you, I didn't notice the mistake.

barracuda156 (approved):

PR #117, opened 2026-05-28
git: fix build on Tiger

ForestExpertise:
make fails to build git on Tiger, so I switched where it was hardcoded in the Portfile to ${build.cmd} and it built. Let me know if that change is problematic.


barracuda156:
I think it’s okay.

PR #120, opened 2026-05-28
claws-mail: Update to 4.4.0

ForestExpertise:
Claws-mail had a new release a couple months ago. It hasn't finished compiling on my Tiger machine, but I figure fixes can be a separate commit for clarity reasons anyway. You may want to test on 10.6 before merging.


barracuda156:
@ForestExpertise Thanks!

PR #133, opened 2026-06-04
sigil-qt4: Fix build on Tiger

ForestExpertise:
Found an elegant, safe, and easy to modify for this situation patch here: https://github.com/macports/macports-ports/commit/6d86a295a5e575ef7f09c5fd3cb8371792f47489


barracuda156 (approved):
As long as the patch is only used on Tiger, I won’t bother verifying if `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` is actually defined in this case. So LGTM.

PR #144, opened 2026-06-08
gobject-introspection: fix Tiger build

ForestExpertise:


barracuda156 (approved):
I can’t test it, but looks fine, as long as it works on your end.

PR #145, opened 2026-06-09
libsdl2-cocoa: fix IOKit Joystick support

doctashay:
Replace the dummy-joystick workaround with a working IOKit driver by fixing a 10.6+ IOHIDDeviceGetService() on 10.5.

Also stops registering the iOS/MFI driver, since we don't have GameController.framework before 10.9.
 
Force feedback/rumble is disabled pre-10.6, but generic HID joystick input now works. Validated with a PS3 and PS4 controller.


barracuda156:
Thank you!

For some I reason was thinking that the upstream code for joystick did not build. I will try this patch.

barracuda156 (approved):
I do not have a joystick to try, but given that you tested in on 10.5, and I confirmed the build works on 10.6, should be good. Thank you!

PR #141, opened 2026-06-07
mupdf: fix build on Leopard by linking libunwind

Matias314:
This resolves:
```
:info:build Undefined symbols for architecture ppc:
:info:build   "__Unwind_Resume", referenced from:
:info:build       _ocr_init in tessocr.o
:info:build ld: symbol(s) not found for architecture ppc
:info:build collect2: error: ld returned 1 exit status
:info:build make: *** [build/shared-release/libmupdf.dylib] Error 1

```

PR #149, opened 2026-06-10
mupdf: Fix Tiger Build

ForestExpertise:


barracuda156 (changes requested):
1. Please use 1.1 version of PG (unless there is some very special reason no to). And add an empty line after PG to please linter.

2. Depending on `binutils` on macOS is untypical. Can’t `cctools` or `darwin-xtools` be used? If not, I am fine with `binutils`, as long as it is only for 10.4, but will be nice if it is avoidable.

ForestExpertise:
I tried cctools first but I couldn't get it to build with that. Darwin-xtools currently doesn't build on Tiger, and looks tedious to fix.
I did switch to 1.1 for legacysupport, I just had tested with 1.0 but I am confident 1.1 will work also.

barracuda156:
Ok, let it be `binutils`. What happened to `tesseract`? It is okay to disable it, just wondering.

ForestExpertise:
Tesseract doesn't build on Tiger currently. I didn't put much effort into fixing it, since it is an optional dependency for mupdf and I don't have much interest in its other dependents.

barracuda156 (approved):

Matias314:
@ForestExpertise my libunwind workaround was to fix a linking issue caused by tesseract OCR: https://github.com/macos-powerpc/powerpc-ports/pull/141. If you are disabling it on Tiger you can probably drop libunwind dependency completely.

ForestExpertise:
> @ForestExpertise my libunwind workaround was to fix a linking issue caused by tesseract OCR: #141. If you are disabling it on Tiger you can probably drop libunwind dependency completely.

Thank you for letting me know. I may test at some point and remove the dependency if it works without it. Perhaps the next time mupdf breaks on Tiger I'll get around to it. Gotta fix gtk4 and some of the other stuff that broke when GNOME did its big update to glib2 and gobject-introspection.

barracuda156:
@ForestExpertise If current gtk4 will be problematic to fix, fallback is likely gonna work. The only port currently requiring recent gtk4 is komikku (via libadwaita). And even it can be built with old gtk4 and libadwaita if my patch is restored (I dropped it since it became unneeded).
(Recent celluloid also needs modern gtk4, but it is broken due to screwed up GL in gtk4, so I keep it at gtk3-based version.)

PR #150, opened 2026-06-10
tdlib: support big endian PPC

doctashay:
tdlib restricts supported platforms to little endian only (i386, x86_64, arm64).

TDLib assumes a little endian host throughout MTProto/TL, so the encrypted auth handshake cannot complete on big endian PPC.
Add ppc/ppc64 and apply a PPC only patch, which no ops elsewhere, and fixes the wire and serialization paths to be endian correct.


barracuda156:
@doctashay Thank you. It will be nice if tdlib is fixed, I will try this.

Could you please drop libsdl2 commit from here (I guess the branch was not rebased to the main)?
And I usually first import existing ports from MacPorts and apply fixes in a follow-up commit (unless something completely trivial). That helps to track down the actual changes without a need to diff ports later on. If it is easier, I can import `tdlib` from MacPorts myself, push that to the main, and you rebase this PR on that.

doctashay:
image

I also have a Cocoa client I am building around tdlib - would rather do this than try to port the official Telegram app or any of the tui clients.

I have no idea how that sdl2 commit got wrapped up in here, but yeah I'll drop it if you can import tdlib from MacPorts and then I can rebase onto that.

Still kinda new to MacPorts, thanks for being patient.

barracuda156:
`tdlib` is imported into PPCPorts now.

barracuda156:
@doctashay Merged in https://github.com/macos-powerpc/powerpc-ports/commit/c1a50d2f5acddfb4fe70b083a6fe6ef55f0569f9
The patch is unchanged; I made a few non-functional changes to portfile.

barracuda156:
@doctashay Follow-up: https://github.com/macos-powerpc/powerpc-ports/commit/8293a1a9510813c4064b480fd29897c5edbbcdf0
This may change C++ ABI for you (depending on how it was built), in which case it will be needed to rebuild the library.

P. S. Did you somehow manage to build `tdlib` with Xcode gcc and system `php` binary?

doctashay:
> @doctashay Follow-up: [8293a1a](https://github.com/macos-powerpc/powerpc-ports/commit/8293a1a9510813c4064b480fd29897c5edbbcdf0) This may change C++ ABI for you (depending on how it was built), in which case it will be needed to rebuild the library.
> 
> P. S. Did you somehow manage to build `tdlib` with Xcode gcc and system `php` binary?

This built fine for me, thanks. No Xcode gcc or system php - I only tested with gcc14.

barracuda156:
But the port will pick Xcode gcc by default on 10.5. As long as either PPCPorts or MacPorts base are used. To change that default something should be done: requiring higher standard, blacklisting Xcode gcc or explicitly setting compiler to a different value.

doctashay:
> But the port will pick Xcode gcc by default on 10.5. As long as either PPCPorts or MacPorts base are used. To change that default something should be done: requiring higher standard, blacklisting Xcode gcc or explicitly setting compiler to a different value.

Assuming I missed this because gcc14 is the default compiler for my PPCPorts install.

barracuda156:
> Assuming I missed this because gcc14 is the default compiler for my PPCPorts install.

So do you build literally everything with it? Say, qt4-mac. How does bootstrap procedure goes, if you start from only Xcode?
(I am interested because I was considering somewhat similar changes in the future.)

PR #152, opened 2026-06-13
gtk4: Fix Tiger build

ForestExpertise:


gnome/gtk4/files/patch-support-tiger.diff
```diff
@@ -0,0 +1,11 @@
+--- gsk/gskpathbuilder.c
++++ gsk/gskpathbuilder.c
+@@ -1706,7 +1706,7 @@
+ gsk_path_builder_add_op (GskPathBuilder         *builder,
+                          GskPathOperation        op,
+                          const graphene_point_t *pts,
+-                         size_t                  n_pts,
++                         unsigned int            n_pts,
```
barracuda156, line 8: What happens on 64-bit Tiger?

ForestExpertise: I don't have a G5, so I have no way of knowing.

barracuda156: Could you say what was the error to begin with?

Not that I care too much about hypothetical 64-bit builds on 10.4, but if possible, we rather make a universally correct fix and not just incidentally working.

ForestExpertise: IMG_20260614_115303394

So the method expects long unsigned int instead of size_t
I can probably hack in a -Wno-error=incompatible-pointer-types flag into meson.build somewhere and that will probably fix it both for ppc and ppc64, though I expect ppc64 will be broken in other ways.

barracuda156: Ok, given that only 10.5 is supported here for ppc64, Tiger-specific patch is fine, I won’t investigate further.

PR #153, opened 2026-06-13
libadwaita: Fix Tiger build

ForestExpertise:


gnome/libadwaita/Portfile
```diff
@@ -45,3 +45,8 @@ depends_lib-append  port:appstream \
 
 # cc1: error: unrecognized command line option "-std=gnu11"
 compiler.c_standard 2011
+
+platform darwin 8 {
+    configure.ldflags-append \
+                    -fstack-protector-strong
```
barracuda156, line 51: Does `-fstack-protector-strong` work as `ldflag`? It seems to be a compiler flag, not linker one: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html

ForestExpertise: This has been tested and works on my Ibook G4. Entirely possible it works as a cflag as well, but ldflag is what I tested. 
The ticket I based my solution on was this:  https://trac.macports.org/ticket/63469 where stack_chk_fail was also a problem during linking on Tiger. They also suggest configure.ldflags-append -lssp which likely also works.
But the only one I tested yesterday was configure.ldflags-append -fstack-protector-strong.
If you would prefer I test and then propose one of the other solutions for this and gtksourceview5 instead, I am open to doing that, but it might be a day or so.

barracuda156: Ok, let it be ldflag, I will merge as is. Not worth spending time to rebuild everything on an iBook.

PR #154, opened 2026-06-13
gtksourceview5: Fix build on Tiger

ForestExpertise:


gnome/gtksourceview5/Portfile
```diff
@@ -80,6 +80,11 @@ if {${universal_possible} && [variant_isset universal]} {
     destroot.env-append    "CC=${configure.cc} ${configure.cc_archflags}"
 }
 
+platform darwin 8 {
+    configure.ldflags-append \
+                    -fstack-protector-strong
```
barracuda156, line 85: Same question as with `libadwaita`. I think this should be a `cflag`, but if you are sure about ld, I will merge these.

barracuda156 (approved):
This likely should rather be `configure.cflags-append -fstack-protector-strong`, but since this works, let it be.

PR #156, opened 2026-06-14
libcmatrix: fix Tiger build

ForestExpertise:
There's something funky about size_t causing errors, but at least in most cases they are errors that can be downgraded to warnings trivially. Gcc10 likely wouldn't error on a lot of this code, but adding some cflags is a small price to pay for a more modern gcc


ForestExpertise:
IMG_20260614_174155149

Screenshot of the error if anyone is curious.

barracuda156:
> but adding some cflags is a small price to pay for a more modern gcc

Yeah, w/e, I am not gonna debug this, and as long as the fix works, good with me.

PR #158, opened 2026-06-15
imgui: add Dear ImGui port

doctashay:
Add a new imgui port for Dear ImGui 1.92.8. The port builds and installs the core static library, public headers, SDL2/OpenGL 2 backend sources, examples and docs.

It includes a PowerPC/big endian color packing patch needed by the OpenGL2 backend, which consumes ImDrawVert colors and RGBA32 font texture data as raw RGBA bytes.

### DIFF
--- a/devel/imgui/Portfile
+++ b/devel/imgui/Portfile
@@ -1,0 +1,88 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        ocornut imgui 1.92.8 v
+revision            0
+categories          devel graphics
+license             MIT
+maintainers         nomaintainer
+
+description         Bloat-free graphical user interface library for C++
+long_description    Dear ImGui is a fast, portable, renderer-agnostic, \
+                    immediate-mode graphical user interface library for C++. \
+                    This port installs a static core library plus the upstream \
+                    headers, backends, examples, fonts, and misc integration \
+                    sources for applications to compile their selected platform \
+                    and renderer backends.
+
+homepage            https://github.com/ocornut/imgui
+
+checksums           rmd160  c2a0f9a38f1df753aef100a71679a051dc47021c \
+                    sha256  fecb33d33930e12ff53a34064e9d3a06c8f7c3e04408f14cd36c80e3faac863b \
+                    size    2081526
+github.tarball_from archive
+
+patchfiles          patch-imgui-big-endian-col32.diff
+
+use_configure       no
+
+compiler.cxx_standard \
+                    2011
+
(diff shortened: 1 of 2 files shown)


barracuda156 (commented):
@doctashay Thank you! Looks good, and we can merge it, but I just wanted to confirm: `imgui` doesn’t have any build system of their own?
It is pretty uncommon to build ports via portfile code itself (I did that in `curl-impersonate`, but that was a very complicated port with badly broken original makefile). But I don’t have objections in principle: if there is nothing from upstream, adding our own makefile would not be less code.

barracuda156:
Besides, is this https://github.com/ocornut/imgui/tree/924f57181b4f7cdde1f11dc25d29e8bb497a90ed/backends omitted on purpose?

I have a port for `implay`, which uses `imgui` (but I did not make a port for the latter, just set it as an extra source), and GL stuff is needed for it.

doctashay:
> `imgui` doesn’t have any build system of their own? 

Upstream's intended model is that whatever project you're working on compiles the core imgui files PLUS whatever backend files needed. So the portfile builds only the platform independent core into `libimgui.a`.

I avoided building the backends into the library because each backend pulls different optional dependencies. Typically your app chooses one platform backend plus one renderer backend, and on PowerPC this is almost always going to be SDL2 plus OpenGL2/modified OpenGL3. 

The backends folder isn't omitted - the headers are installed in `include/imgui/backends`,  and the complete upstream directory is copied under `share/imgui/backends`. 

If you prefer the backend source files beside the headers, I can just adjust the destroot phase to install the files there too - just let me know.

Typically something like imgui wouldn't end up in a package manager like this, but since we're on big endian PowerPC, having this port is useful for centralizing those fixes and gives us a good copy to depend on. Many game ports and 3d applications depend on ImGui so I think this is a good approach.

barracuda156:
I am certainly in favor of having a port rather than pulling in the sources every time for every dependent.

Would you be interested to get `implay` working? Unrelated to this PR as such, but if it works, then we get a GLFW-based player. I got it working on Catalina, but on PowerPC the app did not run. I guess something with GL not working yet.

barracuda156 (approved):
Let me merge it first, and then whatever needed can be adjusted. This will be cleaner anyway.

doctashay:
> I am certainly in favor of having a port rather than pulling in the sources every time for every dependent.
> 
> Would you be interested to get `implay` working? Unrelated to this PR as such, but if it works, then we get a GLFW-based player. I got it working on Catalina, but on PowerPC the app did not run. I guess something with GL not working yet.

Unrelated but I did get ImPlay building - unfortunately it only boots in X11 currently and is very crashy/doesn't work right. I'll keep working on this and hopefully have a port to upload soon.

barracuda156:
@doctashay Thank you very much, this will be very nice.

To avoid X11 it should use GLFW (and GLFW use native backend). My GLFW port has some fixes for 10.6 which makes it usable (simple-gl-viewer works, for example), but those are not enough to support video, it seems. Getting GLFW properly working will give us an alternative to SDL.

PR #160, opened 2026-06-17
youtube-viewer: allow other mpv ports to satisfy dependency.

ForestExpertise:
This change allows ports like mpv-legacy to satisfy the dependency.


barracuda156 (approved):
Thank you, this is useful

Issue #131, opened 2026-06-01
`ffmpeg/ffmpeg8 +x11` transitively pulls in a lot of extra deps and creates dependency cycle (due to `opencolorio +x11`)


barracuda156:
While normally `ffmpeg*` ports do not default to `+x11`, and `ffmpeg +opencolorio` works fine, if some port requires `ffmpeg +x11`, installing from scratch breaks down due to a circular dependency (and potentially broken `openimageio` port plus a lot of extra deps):

```
% port deps opencolorio
Full Name: opencolorio @2.5.1_0+python314
Build Dependencies:   cmake, pkgconfig, py314-pybind11, boost181
Library Dependencies: expat, imath, minizip-ng, pystring, yaml-cpp, zlib, python314

% port deps opencolorio +x11
Full Name: opencolorio @2.5.1_0+python314+x11
Build Dependencies:   cmake, pkgconfig, py314-pybind11, boost181
Library Dependencies: expat, imath, minizip-ng, pystring, yaml-cpp, zlib, glew, lcms2, openimageio,
                      python314, freeglut, libGLU, mesa
% port deps openimageio
Full Name: openimageio @2.1.20.0_17+python314
Build Dependencies:   cmake, robin-map, pybind11
Library Dependencies: field3d, openexr2, squish, tbb, ffmpeg, libfmt9, zlib, bzip2, libpng, tiff,
                      libjpeg-turbo, openjpeg, opencolorio, opencv4, freetype, libheif, ptex, libraw,
                      dcmtk, python314, py314-numpy, boost176, qt64-qtbase
```
(Q6 is not pulled in for powerpc, of course, but still, depending on `openimageio` breaks things down.)

While we could just disable `+opencolorio` in ffmpeg by default, I already have to build multiple ffmpeg variants for each port, and adding yet another is bothersome.

Need to think what is the optimal solution here.

Also related: https://github.com/macos-powerpc/powerpc-ports/issues/17

ForestExpertise:
For Tiger, we may as well disable opencolorio as a default variant, as I don't recall it building. About to make a PR

PR #161, opened 2026-06-17
ffmpeg8: remove opencolorio as a default variant on Tiger

ForestExpertise:
Fixes circular dependency, and I don't recall opencolorio building anyway. If someone fixes opencolorio, adding the variant back in is trivial.

### DIFF
--- a/multimedia/ffmpeg8/Portfile
+++ b/multimedia/ffmpeg8/Portfile
@@ -320,8 +320,6 @@ platform darwin {
 
         configure.args-append \
                     --disable-asm
-
-        default_variants-append -opencolorio
     }
 }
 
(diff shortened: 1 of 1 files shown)


barracuda156:
> Fixes circular dependency

There should be none, unless you require ffmpeg +x11 specifically (which happens in a few instances, but those need some manual procedure of ports activation anyway). Related: https://github.com/macos-powerpc/powerpc-ports/issues/131

But there is no point in requiring opencolorio on Tiger, of course.

multimedia/ffmpeg8/Portfile
```diff
@@ -320,6 +320,8 @@ platform darwin {
 
         configure.args-append \
                     --disable-asm
+
+        default_variants-append -opencolorio
```
barracuda156: Please don’t do it this way (even if it works, it is against the logic).

Just wrap `default_variants-append +opencolorio` in a condition that it is either not darwin or os.major > 8.

barracuda156 (changes requested):

multimedia/ffmpeg8/Portfile
ForestExpertise: I have changed it to what I believe you prefer, but funny enough the other way was what I tested and works fine for disabling the variant.
But I can see the appeal of handling it the other way.

barracuda156: I get that `+1 –1` equals to 0, but this seems unnecessary to do it that way :)

barracuda156 (approved):
While this disables it for Linux as well, it was not tested on Linux to begin with, so does not matter. Once I have time for that, I can adjust the condition.

PR #167, opened 2026-06-29
openjdk8-powerpc: add ppc64 server JIT support

doctashay:
Add tested Darwin/PPC64 server VM support for OpenJDK 8, enabling the JIT by default on ppc64 while keeping ppc on Zero. Includes HotSpot runtime fixes validated against interpreter, mixed JIT, and forced JIT benchmark runs.

Validated with DaCapo 9.12 small suite: luindex, lusearch, pmd, and xalan all pass under -Xint, mixed JIT, and -Xbatch on PPC64 Darwin. Mixed JIT showed roughly 1.7x to 23x speedup over interpreter depending on workload.

Primary motivation for this patch was to improve performance on my RuneScape private server.

### DIFF
--- a/java/openjdk8-powerpc/Portfile
+++ b/java/openjdk8-powerpc/Portfile
@@ -70,7 +70,8 @@ platform powerpc {
                     1015-Some-hacks-to-fix-the-build-for-powerpc.patch \
                     1016-9bd9b56304528ad1ab0a0ee6e19877f313d391e0.patch \
                     1017-Revert-8374899-8u-Fully-handle-clang-as-the-toolchai.patch \
-                    1018-Allow-system-freetype.patch
+                    1018-Allow-system-freetype.patch \
+                    1019-Darwin-PPC64-server-JIT-support.patch
 }
 
 # Active libnet breaks linking of libnio which should link
@@ -265,7 +266,11 @@ if {![variant_isset debug] && ![variant_isset release]} {
     default_variants-append +release
 }
 
-default_variants-append     +zero
+if {${configure.build_arch} eq "ppc64"} {
+    default_variants-append +server
+} else {
+    default_variants-append +zero
+}
 
 build.type          gnu
 build.target        images
--- a/java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
+++ b/java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
@@ -1,0 +1,2132 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Codex <codex@local>
+Date: Mon, 29 Jun 2026 05:30:00 -0600
+Subject: [PATCH] Darwin/PPC64 server JIT support
+
+Add the BSD/PPC os_cpu support used by the server VM and apply the
+Darwin/PPC64 HotSpot fixes validated with DaCapo across -Xint, mixed,
+and -Xbatch modes.
+
+---
+ hotspot/src/cpu/ppc/vm/c2_globals_ppc.hpp       | 9 ++++++++-
+ hotspot/src/cpu/ppc/vm/macroAssembler_ppc.cpp  | 8 ++++----
+ hotspot/src/cpu/ppc/vm/ppc.ad                  | 30 +++++++++++++++----------
+ hotspot/src/os/bsd/vm/os_bsd.cpp               | 28 ++++++++++++++++++++++-
+ hotspot/src/share/vm/opto/parse2.cpp           | 4 ++++
+ 5 files changed, 59 insertions(+), 20 deletions(-)
+
+diff --git hotspot/src/cpu/ppc/vm/c2_globals_ppc.hpp hotspot/src/cpu/ppc/vm/c2_globals_ppc.hpp
+--- hotspot/src/cpu/ppc/vm/c2_globals_ppc.hpp
++++ hotspot/src/cpu/ppc/vm/c2_globals_ppc.hpp
+@@ -52,7 +52,14 @@ define_pd_global(intx, INTPRESSURE,                  25);
+ define_pd_global(intx, InteriorEntryAlignment,       16);
+ define_pd_global(intx, NewSizeThreadIncrease,        ScaleForWordSize(4*K));
+ define_pd_global(intx, RegisterCostAreaRatio,        16000);
++#if defined(__APPLE__) && defined(_LP64)
++// The Darwin/PPC64 TLAB fast path is not reliable in this port yet. DaCapo
++// allocation-heavy workloads abort in object sizing/forwarding with TLABs on,
++// while the same interpreter and JIT modes pass with -XX:-UseTLAB.
++define_pd_global(bool, UseTLAB,                      false);
++#else
+ define_pd_global(bool, UseTLAB,                      true);
++#endif
+ define_pd_global(bool, ResizeTLAB,                   true);
+ define_pd_global(intx, LoopUnrollLimit,              60);
+ 
+diff --git hotspot/src/cpu/ppc/vm/macroAssembler_ppc.cpp hotspot/src/cpu/ppc/vm/macroAssembler_ppc.cpp
+--- hotspot/src/cpu/ppc/vm/macroAssembler_ppc.cpp
++++ hotspot/src/cpu/ppc/vm/macroAssembler_ppc.cpp
+@@ -929,7 +929,7 @@ void MacroAssembler::pop_frame() {
+   ld(R1_SP, _abi(callers_sp), R1_SP);
+ }
+ 
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+ address MacroAssembler::branch_to(Register r_function_entry, bool and_link) {
+   // TODO(asmundak): make sure the caller uses R12 as function descriptor
+   // most of the times.
+@@ -1113,7 +1113,7 @@ address MacroAssembler::call_c_using_toc(const FunctionDescriptor* fd,
+   }
+   return _last_calls_return_pc;
+ }
+-#endif // ABI_ELFv2
++#endif // ABI_ELFv2 || PPC64_DIRECT_C_FUNCTIONS
+ 
+ void MacroAssembler::call_VM_base(Register oop_result,
+                                   Register last_java_sp,
+@@ -1128,7 +1128,7 @@ void MacroAssembler::call_VM_base(Register oop_result,
+ 
+   // ARG1 must hold thread address.
+   mr(R3_ARG1, R16_thread);
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+   address return_pc = call_c(entry_point, relocInfo::none);
+ #else
+   address return_pc = call_c((FunctionDescriptor*)entry_point, relocInfo::none);
+@@ -1153,7 +1153,7 @@ void MacroAssembler::call_VM_base(Register oop_result,
+ 
+ void MacroAssembler::call_VM_leaf_base(address entry_point) {
+   BLOCK_COMMENT("call_VM_leaf {");
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+   call_c(entry_point, relocInfo::none);
+ #else
+   call_c(CAST_FROM_FN_PTR(FunctionDescriptor*, entry_point), relocInfo::none);
+
+diff --git hotspot/src/cpu/ppc/vm/ppc.ad hotspot/src/cpu/ppc/vm/ppc.ad
+--- hotspot/src/cpu/ppc/vm/ppc.ad
++++ hotspot/src/cpu/ppc/vm/ppc.ad
+@@ -1015,7 +1015,7 @@ int MachCallDynamicJavaNode::ret_addr_offset() {
+ }
+ 
+ int MachCallRuntimeNode::ret_addr_offset() {
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+   return 28;
+ #else
+   return 40;
+@@ -3739,7 +3739,7 @@ encode %{
+     MacroAssembler _masm(&cbuf);
+     const address start_pc = __ pc();
+ 
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+     address entry= !($meth$$method) ? NULL : (address)$meth$$method;
+     __ call_c(entry, relocInfo::runtime_call_type);
+ #else
+@@ -3772,7 +3772,7 @@ encode %{
+   // Postalloc expand emitter for runtime leaf calls.
+   enc_class postalloc_expand_java_to_runtime_call(method meth, iRegLdst toc) %{
+     loadConLNodesTuple loadConLNodes_Entry;
+-#if defined(ABI_ELFv2)
++#if defined(ABI_ELFv2) || defined(PPC64_DIRECT_C_FUNCTIONS)
+     jlong entry_address = (jlong) this->entry_point();
+     assert(entry_address, "need address here");
+     loadConLNodes_Entry = loadConLNodesTuple_create(C, ra_, n_toc, new (C) immLOper(entry_address),
+@@ -3808,7 +3808,7 @@ encode %{
+     // Create nodes and operands for loading the Toc point.
+     loadConLNodes_Toc = loadConLNodesTuple_create(C, ra_, n_toc, new (C) immLOper((jlong) fd->toc()),
+                                                   OptoReg::Name(R2_H_num), OptoReg::Name(R2_num));
+-#endif // ABI_ELFv2
++#endif // ABI_ELFv2 || PPC64_DIRECT_C_FUNCTIONS
+     // mtctr node
+     MachNode *mtctr = new (C) CallLeafDirect_mtctrNode();
+ 
+@@ -3849,7 +3849,7 @@ encode %{
+     // These must be reqired edges, as the registers are live up to
+     // the call. Else the constants are handled as kills.
+     call->add_req(mtctr);
+-#if !defined(ABI_ELFv2)
++#if !defined(ABI_ELFv2) && !defined(PPC64_DIRECT_C_FUNCTIONS)
+     call->add_req(loadConLNodes_Env._last);
+     call->add_req(loadConLNodes_Toc._last);
+ #endif
+@@ -3865,7 +3865,7 @@ encode %{
+     // Insert the new nodes.
+     if (loadConLNodes_Entry._large_hi) nodes->push(loadConLNodes_Entry._large_hi);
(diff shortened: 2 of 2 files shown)


java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
+-    address mem_serialize_page = (address) ::mmap(NULL, Bsd::page_size(), PROT_READ | PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
++    address mem_serialize_page = (address)MAP_FAILED;
++#if defined(__APPLE__) && defined(PPC) && defined(_LP64)
++    // Darwin/PPC64 may allocate this early page in the low dylib area before
++    // later JDK libraries are loaded. Use a high hint so dyld cannot remap the
++    // serialize page out from under the VM.
++    const uintptr_t serialize_page_hint = 0x200000000ULL;
++    mem_serialize_page = (address) ::mmap((void*)serialize_page_hint, Bsd::page_size(),
++                                          PROT_READ | PROT_WRITE,
++                                          MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
```
barracuda156, line 232: `MAP_ANONYMOUS` is not defined on relevant to powerpc macOS versions. We can either use `MAP_ANON` here directly or define `MAP_ANONYMOUS` to `MAP_ANON` whenever the former is undefined. (`legacy-support` has that, but better avoid an extra dependency just for this trivial thing.)

Or is this already added elsewhere?

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++sysdefs = -D_ALLBSD_SOURCE -D_GNU_SOURCE -DPPC
+
+diff --git hotspot/make/bsd/platform_ppc64 hotspot/make/bsd/platform_ppc64
+new file mode 100644
+--- /dev/null
++++ hotspot/make/bsd/platform_ppc64
+@@ -0,0 +1,17 @@
++os_family = bsd
++
++arch = ppc
```
barracuda156, line 285: Is this okay to use `ppc` here and not `ppc64`? I don’t remember now what is done with these values. Of course, we do not want something to smuggle `-arch ppc` on ppc64 build, even if compilation works (say, building a FAT object can work, but should not be done without `+universal`).

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++    return;
++#endif
+     d = array_addressing(T_OBJECT, 1);
+     if (stopped())  return;     // guaranteed null or range check
+     array_store_check();
+
+diff --git hotspot/make/bsd/platform_ppc hotspot/make/bsd/platform_ppc
+new file mode 100644
+--- /dev/null
++++ hotspot/make/bsd/platform_ppc
```
barracuda156, line 258: But for 32-bit ppc the build is _not_ changed (yet)?

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++# Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
++#
++# Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
++# or visit www.oracle.com if you need additional information or have any
++# questions.
++#
++#
++
++# make c code know it is on a 64 bit platform.
++CFLAGS += -D_LP64=1
```
barracuda156, line 332: Doesn’t gcc define `__LP64__` when building for a 64-bit arch?

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++ifeq ($(filter $(OPENJDK_TARGET_CPU_ENDIAN),big little),)
++  $(error OPENJDK_TARGET_CPU_ENDIAN value should be 'big' or 'little')
++endif
++
++ifeq ($(OPENJDK_TARGET_CPU_ENDIAN),big)
++  # fixes `relocation truncated to fit' error for gcc 4.1.
++  CFLAGS += -mminimal-toc
++
++  # finds use ppc64 instructions, but schedule for power5
++  CFLAGS += -mcpu=powerpc64 -mtune=power5 -minsert-sched-nops=regroup_exact -mno-multiple -mno-string
```
barracuda156, line 349: Any special reason to use `-mtune=power5`? G5 is _not_ power5. Also, there is `-mtune=G5` / `-mtune=970`, and we know this is the only CPU that can possibly run powerpc64-darwin.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++  OPENJDK_TARGET_CPU_ENDIAN:=$(if $(filter ppc64le,$(shell uname -m)),little,big)
++endif
++
++ifeq ($(filter $(OPENJDK_TARGET_CPU_ENDIAN),big little),)
++  $(error OPENJDK_TARGET_CPU_ENDIAN value should be 'big' or 'little')
++endif
++
++ifeq ($(OPENJDK_TARGET_CPU_ENDIAN),big)
++  # fixes `relocation truncated to fit' error for gcc 4.1.
++  CFLAGS += -mminimal-toc
```
barracuda156, line 346: Is this flag accepted by modern gcc versions? If not, it should be patched in conditionally. (I know we use Apple gcc now, but it still makes sense to take care not to introduce potential issues for modern compilers, since we may need to switch to those later, or for a newer JDK.)

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++
++#ifndef OS_CPU_BSD_PPC_VM_ATOMIC_BSD_PPC_INLINE_HPP
++#define OS_CPU_BSD_PPC_VM_ATOMIC_BSD_PPC_INLINE_HPP
++
++#include "runtime/atomic.hpp"
++#include "runtime/os.hpp"
++#include "vm_version_ppc.hpp"
++
++#ifndef PPC64
++#error "Atomic currently only implemented for PPC64"
```
barracuda156, line 407: Is there some reason not to use gcc built-ins? Xcode gcc only has old `__sync` built-ins, which present a problem for 32-bit ppc, but on 64-bit those should work fine. (Modern gcc has `__atomic` ones which work for either, but signatures need to be adapted as compared to old built-ins.)

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++ */
++
++#ifndef OS_CPU_BSD_PPC_VM_BYTES_BSD_PPC_INLINE_HPP
++#define OS_CPU_BSD_PPC_VM_BYTES_BSD_PPC_INLINE_HPP
++
++#if defined(VM_LITTLE_ENDIAN)
++#include <byteswap.h>
++
++// Efficient swapping of data bytes from Java byte
++// ordering to native byte ordering and vice versa.
```
barracuda156, line 812: For BE we don’t need any?

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++//                           Load|Store
++// - acquire         orders  Load|Store,       (maps to lwsync)
++//                           Load|Load
++// - fence           orders Store|Store,       (maps to sync)
++//                           Load|Store,
++//                           Load|Load,
++//                          Store|Load
++//
++
++#define inlasm_sync()     __asm__ __volatile__ ("sync"   : : : "memory");
```
barracuda156, line 949: _Probably_ these are provided by `ppc_intrinsics.h`. We have it in our gcc https://github.com/macos-powerpc/powerpc-ports/blob/f24237268895c848d177748ca88861dcde53c4f0/lang/gcc16/files/0001-gcc-config-restore-ppc_intrinsics.h.patch but I think it is borrowed from Xcode gcc anyway.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++# include <poll.h>
++# include <ucontext.h>
++
++// Darwin/PowerPC exposes register state through uc_mcontext->__ss. In LP64
++// signal handlers the third argument is a ucontext64_t; mcontext_t remains the
++// 32-bit PPC context type, so using uc_mcontext decodes the wrong layout.
++#ifdef __APPLE__
++# if defined(_LP64)
++#  define PPC_CONTEXT(uc, r) (((ucontext64_t*)(uc))->uc_mcontext64->__ss.__##r)
++# elif __DARWIN_UNIX03 && (MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_5)
```
barracuda156, line 1123: I think `__DARWIN_UNIX03` is enough, it _is_ defined on 10.5+. On the other hand, using version macros needs to ensure those are actually defined (i.e. the header is pulled in).

P. S. It is safer to use numeric macros, i.e. `1050` in place of `MAC_OS_X_VERSION_10_5`.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++# include <sys/utsname.h>
++# include <sys/socket.h>
++# include <sys/wait.h>
++# include <pwd.h>
++# include <poll.h>
++# include <ucontext.h>
++
++// Darwin/PowerPC exposes register state through uc_mcontext->__ss. In LP64
++// signal handlers the third argument is a ucontext64_t; mcontext_t remains the
++// 32-bit PPC context type, so using uc_mcontext decodes the wrong layout.
```
barracuda156, line 1119: Interesting point, something I do not know. Is there a reference to more details?

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++
++    // Check to see if we caught the safepoint code in the
++    // process of write protecting the memory serialization page.
++    // It write enables the page immediately after protecting it
++    // so we can just return to retry the write.
++    //
++    // The PPC instruction decoder used here is implemented only for Linux
++    // ucontext layouts. On BSD/Darwin, safepoint polling is handled above and
++    // this fallback must not call into the Linux-only decoder.
++#if defined(LINUX)
```
barracuda156, line 1493: Is this defined somehow? A standard macro (I think) us `__linux__`.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++
++#ifndef OS_CPU_BSD_PPC_VM_PREFETCH_BSD_PPC_INLINE_HPP
++#define OS_CPU_BSD_PPC_VM_PREFETCH_BSD_PPC_INLINE_HPP
++
++#include "runtime/prefetch.hpp"
++
++
++inline void Prefetch::read(void *loc, intx interval) {
++  __asm__ __volatile__ (
++    "   dcbt   0, %0       \n"
```
barracuda156, line 1772: Here and elsewhere: does this GNUish syntax work for Darwin? I.e. registers without letter prefices? (It won’t work in plain assembler, but I keep forgetting about inline stuff.)

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++#include "runtime/os.hpp"
++#include "runtime/thread.hpp"
++
++#include <ucontext.h>
++
++#ifdef __APPLE__
++# if __DARWIN_UNIX03 && (MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_5)
++#  define PPC_CONTEXT(uc, r) ((uc)->uc_mcontext->__ss.__##r)
++# else
++#  define PPC_CONTEXT(uc, r) ((uc)->uc_mcontext->ss.r)
```
barracuda156, line 1831: There is a claim above that the structure should be a bit different on 32-bit? What about here? Or statement above is inaccurate?
(Sorry if I miss something, I am doing something else and only have this much of time to browse through the changes.)

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++
++#include "precompiled.hpp"
++#include "runtime/frame.inline.hpp"
++#include "runtime/os.hpp"
++#include "runtime/thread.hpp"
++
++#include <ucontext.h>
++
++#ifdef __APPLE__
++# if __DARWIN_UNIX03 && (MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_5)
```
barracuda156, line 1828: Probably `__DARWIN_UNIX03` suffices, see above.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
```diff
@@ -0,0 +1,2124 @@
...
++ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
++ * or visit www.oracle.com if you need additional information or have any
++ * questions.
++ *
++ */
++
++#ifndef OS_CPU_BSD_PPC_VM_THREADLS_BSD_PPC_HPP
++#define OS_CPU_BSD_PPC_VM_THREADLS_BSD_PPC_HPP
++
++  // Processor dependent parts of ThreadLocalStorage
```
barracuda156, line 2063: TLS is not supported by Xcode gcc or our macOS natively, though we do get emulated TLS in modern gcc (when it is enabled).

What does JDK expect here?

java/openjdk8-powerpc/Portfile
```diff
@@ -70,7 +70,8 @@ platform powerpc {
                     1015-Some-hacks-to-fix-the-build-for-powerpc.patch \
                     1016-9bd9b56304528ad1ab0a0ee6e19877f313d391e0.patch \
                     1017-Revert-8374899-8u-Fully-handle-clang-as-the-toolchai.patch \
-                    1018-Allow-system-freetype.patch
+                    1018-Allow-system-freetype.patch \
+                    1019-Darwin-PPC64-server-JIT-support.patch
```
barracuda156, line 74: As long as ppc 32-bit is not switched to a native implementation, better apply the patch conditionally. Just as good for ppc64, but strictly safer for ppc32 (I mean, 2k LoC…).

barracuda156 (changes requested):
@doctashay This is pretty cool if this works. Thank you for the PR.

I added some comments, I think a few minor things may have to be adjusted, and there are a couple of questions.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: The formatting of this patch file got messed up and some content was left out. I am going to close this PR and re-open it shortly.

barracuda156: Ok sure

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: > Doesn’t gcc define `__LP64__` when building for a 64-bit arch?

Yes, but HotSpot checks `_LP64` in several places, so the makefile just defines `_LP64` explicitly for the VM build.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: Yeah, this compiled with the toolchain I used for the test build. It's GCC inline asm rather than plain assembler, so the operand substitution is what handles the register formatting here.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: Agreed, thanks.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: My understanding is this is HotSpot's own thread local lookup layer, not compiler TLS like `__thread`. The VM uses ThreadLocalStorage to find the current HotSpot thread object, so this should be going through HotSpot's TLS setup. It compiled and ran in the build I tested, but I can adjust it if there's something I'm missing...

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: I believe this is okay but I'll double check it. HotSpot seems to use `arch` as the CPU family, so this stays `ppc`, and `arch_model` / `lib_arch` carry the 64-bit distinctions. I gated the patch to `configure.build_arch eq "ppc64"`  so the 32-bit build shouldnt be entering this path.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
doctashay: This was just driven by testing at runtime so I don't have a specific doc to point to. In the `LP64` signal handler path, reading the third signal argument as plain `ucontext_t` gave the wrong register layout. Treating it as `ucontext64_t` fixed PC/register extraction for the JIT signal handling path.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
barracuda156: This is fine, I just asked. As long as it doesn’t look unreasonable and is tested, all good.

java/openjdk8-powerpc/files/1019-Darwin-PPC64-server-JIT-support.patch
barracuda156: I am more concerned about 32-bit case handling. If it was untested (I guess it wasn’t?), rather use conventional structure from system header, consistently.

PR #169, opened 2026-07-06
gcc16: Fix Tiger build

ForestExpertise:
Here it is. I may consult with gcc upstream about the fortran and incompatible pointer type errors, but I used this to rebuild a lot of ports over the last weekend and everything seems to be working fine. I think any further changes can wait until gcc 16.2

### DIFF
--- a/lang/gcc16/Portfile
+++ b/lang/gcc16/Portfile
@@ -262,14 +262,14 @@ platform darwin 8 {
                 port:flex \
                 port:gmake
 
-    # gmake dependency may be able to be dropped next release, problem fixed in commit b081225
+    # gmake dependency may be able to be dropped next release, problem fixed in https://github.com/gcc-mirror/gcc/commit/b081225
     build.cmd ${prefix}/bin/gmake
 
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117834
     patchfiles-append   darwin8-define-PTHREAD_RWLOCK_INITIALIZER.patch
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117857
     patchfiles-append   darwin8-ttyname_r.patch
-    # issue fixed in commit 83f0ad4, was also added later in a commit for gcc16, can be dropped next release
+    # issue fixed https://github.com/gcc-mirror/gcc/commit/83f0ad40a472d98bf1ed38b7d1c4865fdcb5e068 can be dropped next release
     patchfiles-append   darwin8-iconv.patch
     # workaround for Tiger not having _SC_NPROCESSORS_ONLN
     # adapted from https://trac.macports.org/ticket/45832
--- a/lang/gcc16/Portfile
+++ b/lang/gcc16/Portfile
@@ -259,12 +259,23 @@ platform darwin {
 # Keep Tiger stuff isolated
 platform darwin 8 {
     depends_build-append \
-                port:flex
+                port:flex \
+                port:gmake
+
+    # gmake dependency may be able to be dropped next release, problem fixed in commit b081225
+    build.cmd ${prefix}/bin/gmake
 
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117834
     patchfiles-append   darwin8-define-PTHREAD_RWLOCK_INITIALIZER.patch
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117857
     patchfiles-append   darwin8-ttyname_r.patch
+    # issue fixed in commit 83f0ad4, was also added later in a commit for gcc16, can be dropped next release
+    patchfiles-append   darwin8-iconv.patch
+    # workaround for Tiger not having _SC_NPROCESSORS_ONLN
+    # adapted from https://trac.macports.org/ticket/45832
+    patchfiles-append   darwin8-core-count.patch
+    # surpress some incompatible-pointer-type errors where needed
+    patchfiles-append   darwin8-pointer-types.patch
 }
 
 configure.env-append \
--- a/lang/gcc16/files/darwin8-core-count.patch
+++ b/lang/gcc16/files/darwin8-core-count.patch
@@ -18,7 +18,7 @@
      num_images_char = getenv ("NUMBER_OF_PROCESSORS");
  #elif defined (__hpux__)
      return mpctl (MPC_GETNUMSPUS, 0, 0);
-+#elif MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050
++#elif defined(__APPLE__) && (MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
 +{
 +    int nm[2];
 +    size_t len = 4;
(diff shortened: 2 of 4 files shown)


lang/gcc16/files/darwin8-core-count.patch
```diff
@@ -0,0 +1,37 @@
+--- libgfortran/caf/shmem/supervisor.c
++++ libgfortran/caf/shmem/supervisor.c
+@@ -31,6 +31,11 @@
+ #include <signal.h>
+ #include <string.h>
+ #include <unistd.h>
++#include "AvailabilityMacros.h"
++#if __MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050
```
barracuda156: These should be `MAC_OS_X_VERSION_MIN_REQUIRED` and `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__`, I believe (I mean, form of macros at least).

lang/gcc16/files/darwin8-core-count.patch
```diff
@@ -0,0 +1,37 @@
...
++#include <sys/sysctl.h>
++#endif
+ #ifdef HAVE_WAIT_H
+ #include <wait.h>
+ #elif HAVE_SYS_WAIT_H
+@@ -74,6 +79,22 @@
+     num_images_char = getenv ("NUMBER_OF_PROCESSORS");
+ #elif defined (__hpux__)
+     return mpctl (MPC_GETNUMSPUS, 0, 0);
++#elif __MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050
```
barracuda156: What happens on later macOS?

lang/gcc16/Portfile
```diff
@@ -259,12 +259,23 @@ platform darwin {
 # Keep Tiger stuff isolated
 platform darwin 8 {
     depends_build-append \
-                port:flex
+                port:flex \
+                port:gmake
+
+    # gmake dependency may be able to be dropped next release, problem fixed in commit b081225
```
barracuda156: Could you please point to the actual commits? It is not obvious to me which repo is referred to.

If something was fixed either in upstream gcc or in Iain’s branch, let’s add patches as they are, including commit info. While it is generally a good practice, with compilers it is important to make everything transparent.

lang/gcc16/Portfile
```diff
@@ -259,12 +259,23 @@ platform darwin {
...
     patchfiles-append   darwin8-define-PTHREAD_RWLOCK_INITIALIZER.patch
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117857
     patchfiles-append   darwin8-ttyname_r.patch
+    # issue fixed in commit 83f0ad4, was also added later in a commit for gcc16, can be dropped next release
+    patchfiles-append   darwin8-iconv.patch
+    # workaround for Tiger not having _SC_NPROCESSORS_ONLN
+    # adapted from https://trac.macports.org/ticket/45832
+    patchfiles-append   darwin8-core-count.patch
+    # surpress some incompatible-pointer-type errors where needed
+    patchfiles-append   darwin8-pointer-types.patch
```
barracuda156, line 278: It will be nice to include a bugzilla ticket for this. (And open it, if none exists.)

This is not a general requirement to raise errors to upstream (though often it makes sense to do), but again, the primary compiler is a very special case.

barracuda156 (changes requested):

lang/gcc16/files/darwin8-core-count.patch
ForestExpertise: Later MacOS never reaches this part of the if statement either way (it falls under an earlier category), and the patch is only applied on Tiger.

lang/gcc16/Portfile
ForestExpertise: https://github.com/gcc-mirror/gcc/commit/b081225 should fix gcc build for make 3.80 - what Tiger ships. I would prefer to use a later gmake version in case there is another bug they didn't catch.
I wouldn't mind using gmake as the default for all ports on Tiger period, but that is a different discussion
This is the one for iconv - https://github.com/gcc-mirror/gcc/commit/83f0ad40a472d98bf1ed38b7d1c4865fdcb5e068
It uses a/ b/ for the files patched, which does not work in our build and doesn't match what is is the other patches. But everything else is substantially the same.
Both of these were backported into gcc16 later, and I can trac down the commits on that branch if preferred.

lang/gcc16/Portfile
ForestExpertise: I plan to raise the issue soon, but it was so close to the end of the compilation I just wanted something working. I will let you know when I have filed the bugzilla ticket.

lang/gcc16/files/darwin8-core-count.patch
ForestExpertise: I suspect it works either way, but I have now changed it to what you recommend. I will note the above is what I successfully built gcc16 with.

lang/gcc16/Portfile
barracuda156: Those `a/ b/` can be just deleted without affecting anything else. (When using git, `--no-prefix` does that for patches.)

lang/gcc16/files/darwin8-core-count.patch
barracuda156: Ok, then it should be

```
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050 
+#include <sys/param.h>
+#include <sys/sysctl.h>
+#endif
+#endif
```
(notice `<>`), and 
```
#elif defined(__APPLE__) && (MAC_OS_X_VERSION_MIN_REQUIRED < 1050 || __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 1050)
```
Since outside of macOS `AvailabilityMacros.h` does not exist and related macros are not defined.

ForestExpertise: Okay, let me know if it looks better now.

lang/gcc16/Portfile
ForestExpertise: That is basically what I did. I may have also modified some whitespace, but that would be nonfunctional.

ForestExpertise:
Is there anything else that needs to be done before this can be merged?

lang/gcc16/files/darwin8-core-count.patch
```diff
@@ -0,0 +1,39 @@
+--- libgfortran/caf/shmem/supervisor.c
++++ libgfortran/caf/shmem/supervisor.c
+@@ -31,6 +31,13 @@
+ #include <signal.h>
+ #include <string.h>
+ #include <unistd.h>
++#ifdef __APPLE__
++#include "AvailabilityMacros.h"
```
barracuda156: System headers must be included using `< >`.

barracuda156 (changes requested):
Please use links to commits in gcc upstream instead of just hashes. Yes, you reach the same commit in result, but copypasting the link is easy while guessing which repo to search for the hash is less so.

barracuda156:
@ForestExpertise Could you plz make final changes so that we merge this?

ForestExpertise:
> @ForestExpertise Could you plz make final changes so that we merge this?

Changes should be made now, sorry for the delay, was in the difficult part of my workweek.

barracuda156 (approved):
P. S. I do not test on 10.4 so can’t comment of these patches, but given they are used _only_ on 10.4, no objections to merging.

barracuda156:
@ForestExpertise Thank you!

PR #170, opened 2026-07-06
libjxl: fix Tiger build

ForestExpertise:


graphics/libjxl/Portfile
```diff
@@ -54,6 +54,13 @@ patchfiles          include_order.patch \
 if {${os.platform} eq "darwin" && ${os.major} <= 15} {
     # No support for sized operator delete
     patchfiles-append   sized-deallocation.patch
+    if {${os.major} <= 8} {
+        configure.args-append \
+                            -DJPEGXL_ENABLE_BENCHMARK=OFF
+        configure.args-replace \
```
barracuda156: Please use `pre_args` here, since that is what the PG sets: https://github.com/macos-powerpc/powerpc-ports/blob/8fdeb2840959431e40bb6c96c731291ae47bcabd/_resources/port1.0/group/cmake-1.1.tcl#L222-L243

graphics/libjxl/Portfile
```diff
@@ -54,6 +54,13 @@ patchfiles          include_order.patch \
 if {${os.platform} eq "darwin" && ${os.major} <= 15} {
     # No support for sized operator delete
     patchfiles-append   sized-deallocation.patch
+    if {${os.major} <= 8} {
```
barracuda156: This is nitpicking, but I would prefer either `< 9` or `== 8`, otherwise it may be interpreted to imply some support for `< 8`, which does not exist in reality.

barracuda156 (changes requested):

graphics/libjxl/Portfile
ForestExpertise: Both this and the above should be addressed now.

barracuda156 (approved):

graphics/libjxl/Portfile
barracuda156: Thank you!

PR #171, opened 2026-07-06
ffmpeg7: fix Tiger build

ForestExpertise:


multimedia/ffmpeg7/Portfile
```diff
@@ -133,6 +133,10 @@ platform darwin {
     if {${os.major} < 11} {
         patchfiles-append   patch-endian-audiotoolbox.diff
     }
+
+    if {${os.major} < 9} {
+        configure.cflags-append -Wno-error=incompatible-pointer-types
```
barracuda156, line 138: The portfile already sets this flag below (lines 324–325) for all systems, 10.4 included. If that does not work for some reason, we should find out why, but adding the same flag twice is not an option.

barracuda156 (changes requested):

ForestExpertise:
So you are right, and it does build fine on Tiger on my other machine. I must have done something to override or delete the flag in the port by accident on the machine that wouldn't build without the flag. Closing this as invalid. Sorry for the mistake.

PR #182, opened 2026-07-27
add tiger intel fix for m4

alex-free:


alex-free:
Just realized it may not be obvious, this is why legacy-support needs to be in the overlay:


    # Tiger linker doesn't support -rexxport_library. While a newer cctools
    # could provide this, that itself needs libunwind-headers, which needs
    # legacy_support.
    post-patch {
        reinplace "s|-Wl,-reexport_library,\$(OSLIBDIR)/lib\$(OSLIBNAME)\$(SOEXT)||" \
            ${worksrcpath}/Makefile
    }

barracuda156:
@alex-free Hi, thank you for the PR. We certainly want `m4` to work on 10.4, even though 10.4 is not officially supported, but a few requests on this:

1. As a general convention, if we have to add a new port from MacPorts, do it in two commits: import as-is MacPorts version, then apply fixes/changes. I know this was not 100% strictly followed, but in most cases it is, and that is important to be able to see clearly what changed, and also for MacPorts upstream to be able to pick our fixes, if they wish to. This is especially important for fundamental ports, so here I do care.
2. Please keep 1 PR per port, unless really necessary to change two ports at once or when some ports just need revision up for 1 port change. (Here we need 2 PRs.)
3. Specifically re `legacy-support`: it is by intent aimed to support legacy systems, so MacPorts upstream may care to have it working even on 10.4 (decisions are usually up to a port maintainer, and @fhgwright thoroughly tests `legacy-support` on old systems). So while we can have a local 10.4-only fix in our fork, I would prefer to get upstream feedback first.

@fhgwright Fred, could you please comment of the matter? I rather not diverge from MacPorts here, as long as you maintain the port, but we got some Tiger users, and while I personally do not care about 10.4 working or not, why not make life a bit easier for a few more users.

alex-free:
> @alex-free Hi, thank you for the PR. We certainly want `m4` to work on 10.4, even though 10.4 is not officially supported, but a few requests on this:
> 
> 
> 
> 1. As a general convention, if we have to add a new port from MacPorts, do it in two commits: import as-is MacPorts version, then apply fixes/changes. I know this was not 100% strictly followed, but in most cases it is, and that is important to be able to see clearly what changed, and also for MacPorts upstream to be able to pick our fixes, if they wish to. This is especially important for fundamental ports, so here I do care.
> 
> 2. Please keep 1 PR per port, unless really necessary to change two ports at once or when some ports just need revision up for 1 port change. (Here we need 2 PRs.)
> 
> 3. Specifically re `legacy-support`: it is by intent aimed to support legacy systems, so MacPorts upstream may care to have it working even on 10.4 (decisions are usually up to a port maintainer, and @fhgwright thoroughly tests `legacy-support` on old systems). So while we can have a local 10.4-only fix in our fork, I would prefer to get upstream feedback first.
> 
> 
> 
> @fhgwright Fred, could you please comment of the matter? I rather not diverge from MacPorts here, as long as you maintain the port, but we got some Tiger users, and while I personally do not care about 10.4 working or not, why not make life a bit easier for a few more users.

There's still a 10.4 ppc workaround for the same m4 issue (header just is a bit different) in the actual m4 source code, it just wasn't done for Intel.

I can redo these however you want. Legacy support seems to be on the right track mentioning the linker issue and wanting Apple-gcc42, but we hit the circular dependency issue with cctools which is what would provide a linker that understands what that flag is. The Apple-gcc42 itself from my understanding does not fix that

barracuda156:
> There's still a 10.4 ppc workaround for the same m4 issue (header just is a bit different) in the actual m4 source code, it just wasn't done for Intel.

`m4` part can be merged without consulting anyone, we already have `m4` port here. I cannot merge that though, since the PR has a different change, to another port.

> I can redo these however you want. Legacy support seems to be on the right track mentioning the linker issue and wanting Apple-gcc42, but we hit the circular dependency issue with cctools which is what would provide a linker that understands what that flag is. The Apple-gcc42 itself from my understanding does not fix that.

Let’s see if Fred is willing to address this in the upstream port. If not, then we can have a local fix (anything that is strictly 10.4-only I could merge more or less freely, if it looks reasonable and does not add 1k LoC to the codebase).

fhgwright:
@alex-free 
> There's still a 10.4 ppc workaround for the same m4 issue (header just is a bit different) in the actual m4 source code, it just wasn't done for Intel.

Yes, I'm aware of the register-name issue.  At one time, `m4` upstream didn't address that at all, but MacPorts provided a patch for it.  Then at some point, `m4` upstream added a fix for the `ppc` case but not the `x86` case, so the MacPorts patch was reduced to the latter.  Later, someone removed the MacPorts patch as part of the general 10.4 ethnic cleansing, though I've reinstated that patch here locally.

Meanwhile, I've fixed the register-naming issue in `legacy-support`, though that code hasn't been released yet (see [https://github.com/macports/macports-legacy-support/commit/f6564eaf0552f520689c6964e3a3ff41fb5d4627](url).  As of the next `legacy-support` release, `m4` will be able to work without patches, but it presently doesn't use `legacy-support`, and adding it just for 10.4 might get pushback from the 10.4 haters.  If it needed it for 10.5 (or later), nobody would notice. :-)

Given the current attitude towards 10.4, I'm not planning to make a new -devel version until there's something significant for >10.4 systems.

> I can redo these however you want. Legacy support seems to be on the right track mentioning the linker issue and wanting Apple-gcc42, but we hit the circular dependency issue with cctools which is what would provide a linker that understands what that flag is. The Apple-gcc42 itself from my understanding does not fix that

Actually, the linker provided by `apple-gcc42` does support `reexport_library`, though it has to undergo a two-stage bootstrap procedure when starting from plain Xcode 2.5.

What I've found attempting to build `legacy-support` with Xcode 2.5 is:

- Everything _compiles_ just fine with gcc 4.0 (except for a couple of spurious shadowing warnings that I've recently fixed here).
- The library and the tests build and run just fine when _statically_ linked, except for a few bogus multiple-definition warnings where OS functions are intentionally shadowed by `legacy-support`.
- The linker doesn't support `reexport_library`, but that only affects the "system" library, which most programs don't use.
- There is at least one other issue with dynamic linking, causing build failures in the tests.

This is all based on Xcode 2.5.  To the extent that I've tested it, it can build and run just fine (including the tests) with the Xcode 3.1.4 version of gcc 4.0, since that has a newer linker.

alex-free:
Thanks for the clarification. Somehow I missed installing it without the bootstrap. This is great that legacy support is still adding in Tiger fixes, and I definitely understand your situation with waiting for a new release.

Perhaps for now @barracuda156 we can merge only the m4 tiger fix, and then remove it whenever legacy support is updated?

fhgwright:
@alex-free 
> Perhaps for now @barracuda156 we can merge only the m4 tiger fix, and then remove it whenever legacy support is updated?

Makes sense.  It won't hurt with the new `legacy-support`; it will simply become unnecessary.

barracuda156 (approved):

barracuda156:
@fhgwright @alex-free Thank you!

PR #184, opened 2026-07-27
fix multimedia/HandBrake name on case sensitive fs

alex-free:
https://github.com/macports/macports-ports/tree/master/multimedia/HandBrake
https://github.com/macos-powerpc/powerpc-ports/tree/main/multimedia/handbrake
this is causing the overlay to fail integration on case-sensitive filesystems


barracuda156:
This should also rename the port itself then, in portfile.

alex-free:
@barracuda156 should be good, this is how macports has it as well

alex-free:
anything else need to be done for this?

multimedia/HandBrake/Portfile
```diff
@@ -4,7 +4,7 @@ PortSystem              1.0
 PortGroup               github 1.0
 
 github.setup            HandBrake HandBrake 1.11.0
-name                    [string tolower ${github.project}]
+name                    HandBrake
```
barracuda156: This is already set via `github.setup`, not needed.

barracuda156 (changes requested):

barracuda156:
@alex-free Now it looks like there is no change at all, only renaming dirs but not the port.

alex-free:
> @alex-free Now it looks like there is no change at all, only renaming dirs but not the port.

What exactly do you need in the portfile I'm confused

barracuda156:
> > @alex-free Now it looks like there is no change at all, only renaming dirs but not the port.
> 
> What exactly do you need in the portfile I'm confused

If you want camel-case, you need to make coherent changes. Portfile has this:
```
github.setup            HandBrake HandBrake 1.11.0
name                    [string tolower ${github.project}]
```
This makes the name to be `handbrake`.

alex-free:
Now matches https://github.com/macports/macports-ports/blob/master/multimedia/HandBrake/Portfile.  I just don't want 2 different handbrake ports due to case-insensitivity. I personally do not like camel case and don't know why one port subgroup (multimedia) went all out on it.

multimedia/HandBrake/Portfile
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+
+github.setup            HandBrake HandBrake 1.11.0
+name                    HandBrake
```
barracuda156: Could you just delete `[string tolower ${github.project}]` here, please? There is no need to set name again, it is done by the PG. And please try avoid adding trailing spaces, I can’t keep checking that.

barracuda156 (changes requested):

barracuda156 (approved):

PR #148, opened 2026-06-09
libunwind: Fix Tiger build with legacysupport

ForestExpertise:
Legacy support fixes this, which is lucky because I was taking a shot in the dark.

PR #185, opened 2026-07-29
Libunwind - PPCPorts added Legacy Support Causes Circular Dependency On Tiger

alex-free:
OK, so this is actually what was causing https://github.com/macos-powerpc/powerpc-ports/pull/182#issuecomment-5094643778

Currently if you use this overlay and try to build completely from scratch (no binary archives) from Xcode 2.5, you run into this proccess.

Build apple-gcc42+bootstrap variant (auto-default if GCC 4.2 is not installed, see:  https://github.com/macports/macports-ports/blob/349aaaaf655f89939b0750d4cbe86dc8da554153/lang/apple-gcc42/Portfile#L191)

Then you need to build apple-gcc42 again, in order to get the linker that @fhgwright was talking about . This requires libunwind-headers subport: https://github.com/macos-powerpc/powerpc-ports/blob/main/devel/libunwind/Portfile

libunwind (not the subport) apparently (haven't finished building it yet to see why) needs legacy_support according to the PPCPorts pull request: https://github.com/macos-powerpc/powerpc-ports/pull/148

The problem is we can't build legacy support until after we have the non-bootstrap variant of apple-gcc42.

Potential solutions:
1) This pull request. We can figure out what out of legacy-support is needed for libunwind, and do patches for it without requiring legacy-support.
2) Make libunwind-headers subport a real port. Keep libunwind as it is (pulling in legacy-support). Change apple-gcc42 dependency for libunwind-headers real port.
3) Maybe there is some way to not enable legacy_support on the subport only? This would be ideal.
4) Ignore it. If you have a binary archive of apple-gcc42 (not bootstrap varient) then this whole mess is skipped. I don't particularly like the idea of not being able to bootstrap the system from xcode2.5 though, so currently my ports tree just removes legacy_support as I am aiming for the first option (https://github.com/alex-free/tigerports-ports/blob/master/devel/libunwind/Portfile)

Just figured I'd open a discussion here. Let me know your thoughts


fhgwright:
The standard MacPorts `apple-gcc42` does not use `legacy-support`.  Perhaps you should track down how that got in there.

alex-free:
> The standard MacPorts `apple-gcc42` does not use `legacy-support`. Perhaps you should track down how that got in there.

Correct, and that works as intended. As mentioned above, the pull request to PPCPorts that added this regression is: https://github.com/macos-powerpc/powerpc-ports/pull/148

barracuda156:
@alex-free So the issue is that `legacy-support` is needed for `libunwind` but must not be pulled in from `libunwind-headers`? That’s easy to fix by doing exactly that.

P. S. I do not test any PRs for Tiger, and that one looked fine to me to merge.

alex-free:
> @alex-free So the issue is that `legacy-support` is needed for `libunwind` but must not be pulled in from `libunwind-headers`? That’s easy to fix by doing exactly that.
> 
> P. S. I do not test any PRs for Tiger, and that one looked fine to me to merge.

Exactly yes

barracuda156:
@alex-free Should be fixed

alex-free:
> @alex-free Should be fixed

By far the best fix. Will let you know when this has been tested

alex-free:
@barracuda156 bootstrapping works! Thx

fhgwright:
Something is still fishy here.  The code for `libunwind` hasn't changed in years, and it never needed `legacy-support` in the past.  What changed?

alex-free:
> Something is still fishy here.  The code for `libunwind` hasn't changed in years, and it never needed `legacy-support` in the past.  What changed?
> 
> 

@ForestExpertise can you chime in here on why you did that pull request? What was legacy support for libunwind (not headers subport) needed for on tiger?

barracuda156:
@alex-free Can you check if it builds for you on 10.4?

@fhgwright Does `libunwind` come in some very common dependency paths? If not, it could have been broken for years, easily. Despite some claims to the contrary, nearly nothing gets tested for legacy systems, and even failures on buildbots are ignored indefinitely. If this is the case for 10.6, would be surprising if it was not for 10.4.

alex-free:
> @alex-free Can you check if it builds for you on 10.4?
> 
> 
> 
> @fhgwright Does `libunwind` come in some very common dependency paths? If not, it could have been broken for years, easily. Despite some claims to the contrary, nearly nothing gets tested for legacy systems, and even failures on buildbots are ignored indefinitely. If this is the case for 10.6, would be surprising if it was not for 10.4.

As soon as I fix gcc16 on i386 tiger I can try building it again. The Mac Pro 1,1 Xeon 2.0ghz dual core dual cpu 12GB RAM is a little tied up with libgcc16 right now.

ForestExpertise:
> > Something is still fishy here.  The code for `libunwind` hasn't changed in years, and it never needed `legacy-support` in the past.  What changed?
> 
> @ForestExpertise can you chime in here on why you did that pull request? What was legacy support for libunwind (not headers subport) needed for on tiger?

IMG_20260729_210049536
The above error happens for libunwind without legacy support. My apologies for creating a circular dependency, I failed to realize libunwind-headers was a sub-port using the same Portfile. Sloppy on my part, and since I had everything installed already the circular dependency never came to my notice. Thank you @barracuda156 for fixing it so it only applies to the needed port.

I have no idea if libunwind used to work for Tiger in the past before my PR which accidentally broke libunwind-headers, I suspect not as while there are over 3000 ports that depend on it, few look familiar

alex-free:
> > > Something is still fishy here.  The code for `libunwind` hasn't changed in years, and it never needed `legacy-support` in the past.  What changed?
> 
> > 
> 
> > @ForestExpertise can you chime in here on why you did that pull request? What was legacy support for libunwind (not headers subport) needed for on tiger?
> 
> 
> 
> IMG_20260729_210049536
> 
> The above error happens for libunwind without legacy support. My apologies for creating a circular dependency, I failed to realize libunwind-headers was a sub-port using the same Portfile. Sloppy on my part, and since I had everything installed already the circular dependency never came to my notice. Thank you @barracuda156 for fixing it so it only applies to the needed port.
> 
> 
> 
> I have no idea if libunwind used to work for Tiger in the past before my PR which accidentally broke libunwind-headers, I suspect not as while there are over 3000 ports that depend on it, few look familiar
> 
> 

That makes sense, threading stuff on tiger vs leopard is limited.

Pretty reassuring not a lot that looks familiar depends on it. That probably is why no one noticed.

barracuda156:
The main function of `libunwind` is to poison builds on modern macOS (really, it links in, and then apps crash on launch).

fhgwright:
@ForestExpertise 
[...]
> The above error happens for libunwind without legacy support. My apologies for creating a circular dependency, I failed to realize libunwind-headers was a sub-port using the same Portfile. Sloppy on my part, and since I had everything installed already the circular dependency never came to my notice. Thank you @barracuda156 for fixing it so it only applies to the needed port.
> I have no idea if libunwind used to work for Tiger in the past before my PR which accidentally broke libunwind-headers, I suspect not as while there are over 3000 ports that depend on it, few look familiar

None of my five 10.4 systems has `libunwind` installed, though they all have _`libunwind-headers`_ installed, which is all that's necessary to build `cctools`.  I've seen claims that `libunwind` doesn't even work properly on any OS <10.6, though I haven't personally verified that.

According to `port deps`, building `libunwind` on 10.4 requires `gcc15` on ppc and `clang-11-bootstrap` on x86, but neither of those compilers actually builds on 10.4.  I have no idea why code that's decades old suddenly thinks it needs a new compiler.

@barracuda156 
> The main function of libunwind is to poison builds on modern macOS (really, it links in, and then apps crash on launch).

It was never intended to be used on recent systems, but some people stuck it in the Portfile unconditionally and never fixed it.  Its ostensible documentation page (see `port gohome libunwind`) doesn't even exist any more.

barracuda156:
> According to `port deps`, building `libunwind` on 10.4 requires `gcc15` on ppc and `clang-11-bootstrap` on x86, but neither of those compilers actually builds on 10.4.

gcc16 should, I guess, though it has some 10.4-specific hacks now, which I did not personally test. No idea about `clang-11-bootstrap` or why 10.4 at all tries to use _any_ clang.

alex-free:
> > According to `port deps`, building `libunwind` on 10.4 requires `gcc15` on ppc and `clang-11-bootstrap` on x86, but neither of those compilers actually builds on 10.4.
> 
> 
> 
> gcc16 should, I guess, though it has some 10.4-specific hacks now, which I did not personally test. No idea about `clang-11-bootstrap` or why 10.4 at all tries to use _any_ clang.

Yea you were right to strip clang out from ppcports-base. Not really the best place to ask this but what is actually required to build gcc16 on tiger? Am I correct that Darwin-xtools would be needed because ld64_97 certainly isn't capable of linking it due to sheer size and amount of code (at least i386) and that's what Intel is trying to use right now.

I guess what I'm trying to ask as a high level overview of bootstrapping to gcc16. I can get up to GCC 10 bootstrap just fine, I also see llvm-7.0 mentioned so I know there's a lot going on

I should probably just do port rdeps gcc16 on PowerPC but maybe you could shine some light friend @barracuda156

barracuda156:
> Not really the best place to ask this but what is actually required to build gcc16 on tiger?

I have no idea, I do not build on 10.4, it is not supported :) 

But for the same reason nothing should have changed since gcc14: you should get identical dependency chain. All toolchain improvements here (aside of gcc update) bypassed Tiger.

> because ld64_97 certainly isn't capable of linking it

Unless you build for ppc64, which you probably don’t, ld64-97 should be fine.

> I also see llvm-7.0 mentioned

Not required by default for most purposes. Not for the toolchain at least. Pulled in from non-default variants and a couple of exotic ports.

ForestExpertise:
> > > According to `port deps`, building `libunwind` on 10.4 requires `gcc15` on ppc and `clang-11-bootstrap` on x86, but neither of those compilers actually builds on 10.4.
> > 
> > 
> > gcc16 should, I guess, though it has some 10.4-specific hacks now, which I did not personally test. No idea about `clang-11-bootstrap` or why 10.4 at all tries to use _any_ clang.
> 
> Yea you were right to strip clang out from ppcports-base. Not really the best place to ask this but what is actually required to build gcc16 on tiger? Am I correct that Darwin-xtools would be needed because ld64_97 certainly isn't capable of linking it due to sheer size and amount of code (at least i386) and that's what Intel is trying to use right now.
> 
> I guess what I'm trying to ask as a high level overview of bootstrapping to gcc16. I can get up to GCC 10 bootstrap just fine, I also see llvm-7.0 mentioned so I know there's a lot going on
> 
> I should probably just do port rdeps gcc16 on PowerPC but maybe you could shine some light friend @barracuda156

Can confirm that ld64-97 is all that is required on 32 bit, which is all that has been tested on Tiger.
Libunwind doesn't strictly need gcc15, if it is showing that it is because your macports-base has that as the default compiler. You can fix that in port configure.tcl as described here: https://forums.macrumors.com/threads/ppcports-the-only-ports-you-ever-need-on-a-powerpc.2462897/page-6?post=34289517#post-34289517
You will want either gcc14 (most tested), gcc16 (most modern, and has worked fine for me), or possibly gcc10 (if you hate adding -Wno-error=incompatible-pointer-types to ports)
Port rdeps gcc16 gives the following 
IMG_20260731_220131998

ForestExpertise:
> @ForestExpertise [...]
> 
> > The above error happens for libunwind without legacy support. My apologies for creating a circular dependency, I failed to realize libunwind-headers was a sub-port using the same Portfile. Sloppy on my part, and since I had everything installed already the circular dependency never came to my notice. Thank you @barracuda156 for fixing it so it only applies to the needed port.
> > I have no idea if libunwind used to work for Tiger in the past before my PR which accidentally broke libunwind-headers, I suspect not as while there are over 3000 ports that depend on it, few look familiar
> 
> None of my five 10.4 systems has `libunwind` installed, though they all have _`libunwind-headers`_ installed, which is all that's necessary to build `cctools`. I've seen claims that `libunwind` doesn't even work properly on any OS <10.6, though I haven't personally verified that.
> 
> According to `port deps`, building `libunwind` on 10.4 requires `gcc15` on ppc and `clang-11-bootstrap` on x86, but neither of those compilers actually builds on 10.4. I have no idea why code that's decades old suddenly thinks it needs a new compiler.
> 
> @barracuda156
> 
> > The main function of libunwind is to poison builds on modern macOS (really, it links in, and then apps crash on launch).
> 
> It was never intended to be used on recent systems, but some people stuck it in the Portfile unconditionally and never fixed it. Its ostensible documentation page (see `port gohome libunwind`) doesn't even exist any more.

Yeah, I only built libunwind to fulfill a dependency requirement, I have no idea if it is actually doing anything.
As for why old code thinks it needs a new compiler, macports periodically changes it's default compiler, but you can take control of that here : https://forums.macrumors.com/threads/ppcports-the-only-ports-you-ever-need-on-a-powerpc.2462897/page-6?post=34289517#post-34289517
On 10.4, the compilers that work from macports/PowerPC ports are apple-gcc-42, gcc7, gcc10, gcc14, gcc16 and possibly gcc-48 (only used for TenFourFox and forks, so I haven't really tested)
I feel gcc16 is the way to go, though a couple of my hacks should be run past upstream to see if they have a better idea.
I did have llvm-7.1.1 building on PowerPC ports at one point. I assume it works but I have never really tested it.
Clang-3.4 builds, but is not reported to work. Only useful for fulfilling dependency requirements when testing ports from the distant macports past.

PR #186, opened 2026-07-29
linrad: powerpc build fixes

Matias314:


science/linrad/Portfile
```diff
@@ -16,19 +16,24 @@ homepage            https://www.sm5bsz.com/linuxdsp/linrad.htm
 fetch.type          svn
 worksrcdir          trunk
 svn.url             https://svn.code.sf.net/p/linrad/code/trunk/
-svn.revision        997
-version             20211204-${svn.revision}
+svn.revision        1002
```
barracuda156, line 19: If you update the port, it should be in the PR title too, don’t sneak it in )

Matias314: Thank you for the heads up, I forgot about that D:

PR #188, opened 2026-08-02
Fix Python 27 bootstrap variant on Tiger

alex-free:
Installation on darwin 8 fails due to it expecting a system-expat that doesn't exist on tiger. This is only happening on the bootstrap variant, I assume most people just install python27 and never ran into this issue


barracuda156 (changes requested):
This does not look correct in two ways:
1. `${os.major}` is only defined for Darwin, but you use it is a block that fires for non-darwin as well.
2. Your change has an effect on non-bootstrap port, and it is unclear if those are intended, needed and right.

Also, do not duplicate an arg, it can be written with less code by passing `--with-system-expat` when needed (FFI part remains generic).

alex-free:
I'm going to retest this because it was 100% failing and I see what you are saying, but I think the conditional is now the proper way

lang/python27/Portfile
```diff
@@ -112,17 +112,16 @@ configure.args-append \
...
-    # Non-darwin may not have a system libexpat or libffi, so we can't use them for the bootstrap port 
+    # Non-darwin may not have a system libexpat or libffi, so we can't use them for the bootstrap port
+        configure.args-append \
+                        --with-system-ffi
+
     # If installing on tiger and requesting +bootstrap variant, can not add --with-system-expat
     # ld: library not found for -lexpat 
     # ImportError: No module named pyexpat
-    if {${os.major} == 8} {
+    if {!(${os.platform} eq "darwin" && ${os.major} == 8)} {
```
barracuda156: Your change still applies for non-bootstrap port, where it _presumably_ (I did not test, but it follows from your own words) is not needed. I don’t know what effect that has and whether it is wanted, but as long as a change is not required, rather not make it without a very good reason.

If the whole point is to fix `python27-bootstrap` on 10.4, nothing should spill over into `python27` build.

barracuda156 (changes requested):

lang/python27/Portfile
alex-free: Correct. I need to verify that python27 was failing before any of my changes in a re-test, because I now think the whole port was broken before this. Initially I was only testing the bootstrap variant and might've mixed this up.

barracuda156: Let me look through the logic and suggest how to do it.

barracuda156: Ok then I wait for your input first.

alex-free:
OK @barracuda156 I was right it was just the -bootstrap variant. This should be good now

barracuda156:
Can you make it exactly two commits (import from MacPorts + change for 10.4)? Or it is easier if I import, and you rebase the PR? I cannot squash commits selectively when merging, but we need to keep import as a separate commit.

alex-free:
> Can you make it exactly two commits (import from MacPorts + change for 10.4)? Or it is easier if I import, and you rebase the PR? I cannot squash commits selectively when merging, but we need to keep import as a separate commit.

Yea give me one sec

alex-free:
ok 2 commits, 1 import 1 changed with correction

barracuda156:
@alex-free This can be merged now, please change commit titles to what we need by convention: `python27: import from MacPorts`, `python27-bootstrap: ${whatever text about a change}`.

Issue #187, opened 2026-08-01
GCC16 Tiger i386: Final Linking Fails With ld: 24-bit pointer diff out of range in unwind info in unwind info from <various executables>


alex-free:
OK, so I've been looking into this for a while. I can get up to gcc10-bootstrap no problem on Intel Tiger. This issue occurs in literally the final steps of linking GCC16, and I am pretty certain that this is Intel specific. My understanding/theories are:

- LD64_97 has a maximum allocation size it can use for unwind data (what that is IDK)
- When compiling with GCC10 bootstrap on Intel, this limit is hit when it is not on PPC due to architecture differences (Intel has larger data).
- Even though passing -g0 (disable debug), -fno-asynchronous-unwind-tables, and disabling everything except c, c++, objc, and objc++ (all of which reduces size of what is linked) gets us closer (way less of these errors, all combined only cc1plus is failing) GCC still compiles with -g and -fasynchronous-unwind-tables regardless of what macports passes to it in certain parts.

I understand this isn't PPC nor is it Snow Leopard or Leopard, but perhaps you have some kind of knowledge of this error. I will continue trying things, and document them here at least.  The full log (ran sudo port -d install gcc16 after a failed compile) is attached

[i386-darwin8-gcc16-build.txt](https://github.com/user-attachments/files/30612534/i386-darwin8-gcc16-build.txt)

@ForestExpertise

barracuda156:
This is probably something never tested by anyone. I never had 10.4 i386, and the error itself is unfamiliar. You may try asking Iain or opening a Bugzilla ticket, though if it is not something obvious for upstream, you may need to build outside of MacPorts (or PPCPorts) to reproduce the same failure in a clean (for the purpose of debugging) env.

barracuda156:
@catap Kirill, just in case, have you ever tried building for old i386 some recent gcc (11+)?

catap:
No, I don't

alex-free:
The best compiler I ever got working on Intel Tiger was when I was trying this last in 2022 with @kencu TigerPorts overlay GCC7 worked. I have this saved from back then:

apple-gcc42                    @5666.3         lang/apple-gcc42
apple-gcc42                    @5666.3         lang/apple-gcc42
autoconf                       @2.71           devel/autoconf
automake                       @1.16.5         devel/automake
bzip2                          @1.0.8          archivers/bzip2
cctools                        @949.0.1        devel/cctools
[...]

PR #191, opened 2026-08-03
Set Libgcc/gcc max to 14 on i386 darwin 8

alex-free:
Due to libgcc16 not currently building on i386 tiger (https://github.com/macos-powerpc/powerpc-ports/issues/187), we should limit libgcc to 14 and maximum gcc to gcc14 on this configuration until it is resolved.

I think this is the cleanest way to support this, let me know your thoughts. I believe this is the correct way to allow GCC to build with matching libgcc14 (this is usually a problem, yes?), but you are much more familiar with toolchains then I. I tried to go off of https://github.com/macos-powerpc/powerpc-ports/commit/cff9f9cf48a7637f9b430f5edaa1053ad5e93c01 , so please let me know if this seems correct.

I'm going to rebuild everything from nothing just to sanity check this as well and report back.

Effectively, intel tiger now requests:

alexs-computer-2:~ alex$ sudo port rdeps gcc14
The following ports are dependencies of gcc14 @14.3.0_0:
  xz
    gettext
      xz-bootstrap
        apple-gcc42
          gcc_select
      ncurses
      gettext-tools-libs
        libiconv
          gperf
        libtextstyle
        gettext-runtime
  texinfo
    help2man
      perl5.34
        db48
        gdbm
          readline
      p5.34-locale-gettext
    libunistring
      autoconf
        m4-bootstrap
      automake
      libtool
        m4
      gmake
        lzip
  gcc10-bootstrap
    gcc7-bootstrap
      cctools
        libunwind-headers
      ld64
        ld64-97
          libmacho-headers
  flex
  gmp
  isl
  libmpc
    mpfr
  zlib
  zstd
    lz4
  libgcc14
alexs-computer-2:~ alex$ 

PowerPC tiger now requests:
The following ports are dependencies of gcc14 @14.3.0_0:
  xz
    gettext
      xz-bootstrap
        apple-gcc42
          gcc_select
      ncurses
      gettext-tools-libs
        libiconv
          gperf
        libtextstyle
        gettext-runtime
  texinfo
    help2man
      perl5.34
        db48
        gdbm
          readline
      p5.34-locale-gettext
    libunistring
      autoconf
        m4-bootstrap
      automake
      libtool
        m4
      gmake
        lzip
  gcc10-bootstrap
    gcc7-bootstrap
      cctools
        libunwind-headers
      ld64
        ld64-97
          libmacho-headers
  flex
  gmp
  isl
  libmpc
    mpfr
  zlib
  zstd
    lz4
  libgcc14
    libgcc
      libgcc16

### DIFF
--- a/_resources/port1.0/compilers/gcc_dependencies.tcl
+++ b/_resources/port1.0/compilers/gcc_dependencies.tcl
@@ -2,7 +2,14 @@
 
 # GCC version providing the primary runtime
 # Note settings here *must* match those in the lang/libgcc port and compilers PG
-set gcc_main_version 16
+# GCC 16 and 14 on all systems except i386 tiger (GCC14 only) due to:
+# https://github.com/macos-powerpc/powerpc-ports/issues/187
+global os.arch
+if {${os.major} == 8 && ${os.arch} eq "i386"} {
+    set gcc_main_version 14
+} else {
+    set gcc_main_version 16
+}
 
 # compiler links against libraries in libgcc\d* and/or libgcc-*
 if {[vercmp ${gcc_version} 4.6] < 0} {
--- a/_resources/port1.0/compilers/gcc_dependencies.tcl
+++ b/_resources/port1.0/compilers/gcc_dependencies.tcl
@@ -4,8 +4,8 @@
 # Note settings here *must* match those in the lang/libgcc port and compilers PG
 # GCC 16 and 14 on all systems except i386 tiger (GCC14 only) due to:
 # https://github.com/macos-powerpc/powerpc-ports/issues/187
-global os.arch
-if {${os.major} == 8 && ${os.arch} eq "i386"} {
+global os.arch os.platform
+if {${os.platform} eq "darwin" && ${os.major} == 8 && ${os.arch} eq "i386"} {
     set gcc_main_version 14
 } else {
     set gcc_main_version 16
--- a/_resources/port1.0/compilers/gcc_compilers.tcl
+++ b/_resources/port1.0/compilers/gcc_compilers.tcl
@@ -3,8 +3,14 @@
 # https://trac.macports.org/ticket/57135
 # https://trac.macports.org/ticket/61636
 
-# GCC 14 on all systems
-lappend compilers macports-gcc-16 macports-gcc-14
+# GCC 16 and 14 on all systems except i386 tiger (GCC14 only) due to:
+# https://github.com/macos-powerpc/powerpc-ports/issues/187
+global os.arch
+if {${os.major} == 8 && ${os.arch} eq "i386"} {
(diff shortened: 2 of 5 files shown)


barracuda156 (commented):
I think the right thing to do is fix gcc16 build, and not introduce hacks for an unsupported config scattered all over core components.

If there is no response from GCC upstream on the matter, this can be reconsidered (perhaps in a different form; something like standalone isolated toolchain for 10.4 is conceivable, but not obscure changes which will require thinking over them every time a port is updated).

_resources/port1.0/compilers/gcc_dependencies.tcl
```diff
@@ -2,7 +2,14 @@
 
 # GCC version providing the primary runtime
 # Note settings here *must* match those in the lang/libgcc port and compilers PG
-set gcc_main_version 16
+# GCC 16 and 14 on all systems except i386 tiger (GCC14 only) due to:
+# https://github.com/macos-powerpc/powerpc-ports/issues/187
+global os.arch
+if {${os.major} == 8 && ${os.arch} eq "i386"} {
```
barracuda156: `${os.major}` exists only for Darwin.

alex-free:
> I think the right thing to do is fix gcc16 build, and not introduce hacks for an unsupported config scattered all over core components.
> 
> If there is no response from GCC upstream on the matter, this can be reconsidered (perhaps in a different form; something like standalone isolated toolchain for 10.4 is conceivable, but not obscure changes which will require thinking over them every time a port is updated).

The way this is implemented everything tiger i386 related can be ignored whenever a libgcc update would happen. Tiger i386 would remain at libgcc14 and then everything else would move to the next version. Moving to a newer libgcc and gcc when the time comes will not break this for tiger, and this wont break anything else. Just move everything up from 16 thats at 16 to 18  for example. By that time though I'm hoping the libgcc16 issue is resolved. I am going to follow up with upstream about the issue and try some more things myself as well.

I understand what your saying, but as it is we are stuck with clang3.x and gcc-4.2.1 without having a working libgcc on intel, because literally nothing else will build except bootstrap gcc10/7 without one we can use. I do want to get it to parity with powerpc, but until that happens I think its better to at least have a viable system.

alex-free:
Another thing to consider is having gcc14 work on Intel allows more ports to be tested faster and fixes implemented while a libgcc16 fix is being figured out. Like forest expertise I only have a G4 right now (G5s are quite pricy rn but I really am trying to get one when possible) so it can only do things so fast.

alex-free:
GCC16 fix available, not needed: https://github.com/macos-powerpc/powerpc-ports/pull/193

barracuda156:
Ah, nice, thank you. Please make a minor adjustment there, and we can merge it.

PR #194, opened 2026-08-05
Blacklist Apple GCC For Brotli

alex-free:
Was reported this was broken in mainline macports on Leopard. Apple GCC can not build this correctly in the way that it wants


archivers/brotli/Portfile
```diff
@@ -24,6 +24,11 @@ checksums           rmd160  949e0c56045b760b854714f45b0912936a9be53a \
                     sha256  816c96e8e8f193b40151dad7e8ff37b1221d019dbcb9c35cd3fadbfe6477dfec \
                     size    646315
 
+# Apple gcc disables -fnested-functions by default. Even if this is enabled with a flag, the
+# next error is ld: -allow_stack_execute option can only be used when linking a main executable.
+# Newer compilers support functions to allow this behavior properly. 
+compiler.blacklist apple-gcc-4.2 gcc-4.0
```
barracuda156: This will prohibit `apple-gcc-4.2` but allow `apple-gcc-4.0`, is this intended?

Since the port builds with `gcc-4.2`, it is also a bit strange (though not impossible) if it fails to build with `apple-gcc-4.2` on same systems.

alex-free: On my system this blocks both and gcc16 is next. The priority is vendor compilers then gcc16. I thought I've seen it before where multiple can be listed there. Using debug build it does use gcc16

barracuda156: Priorities may differ across base versions and may change in the future (including by accident). I do not want to have a blacklist which every time gonna make someone puzzled why it exists in such a peculiar form. If the build is confirmed to be broken with these two compilers, I suggest to assume `apple-gcc-4.0` will not work either (unless, of course, that was tested and shown to work fine – but then it must be reflected in comments in portfile).
Non-obvious blacklists should rather have an explanation attached, why this choice was made and which system the breakage has been observed on. Otherwise someone later decides that this was a mistake, and `gcc-4.2` should also be blacklisted, for example.

If all this is about Tiger, then put the whole thing inside `darwin 8`, please. Nothing should be changed for 10.5–10.6 blindly, based on 10.4 quirks.

alex-free: I am blacklisting gcc-4.0 here though. It seems gcc-4.0 can be built in macports and get the apple-gcc-4.0 name however, so I'll add that. I also do have an explanation, I'm blacklisting both.  MacPorts says multiple can be listed here in compiler.blacklist https://trac.macports.org/wiki/CompilerSelection and they are derived in portconfigure its referenced as gcc-4.0: https://github.com/macos-powerpc/ppcports-base/blob/master/src/port1.0/portconfigure.tcl but I also see where it _could_ be apple-gcc-4.0, so give me a second I'll add that and wrap it in darwin8 conditional

barracuda156: N/m, it was either removed or I confused smth. Anyway, `*gcc-4.0` makes everyone satisfied, with fewer characters :)

barracuda156: I could not quickly find refs to `apple-gcc-4.0`, but apparently `llvm-gcc-4.0` was a thing: https://github.com/llvm/llvm-archive/blob/main/llvm-gcc-4.0
So just use `*gcc-4.0`, and we don’t need to worry that some obscure 4.0 creeps in.

alex-free: For sure. Still want the darwin8 conditional?

barracuda156: Strictly speaking this remains untested, but I am not gonna build `apple-gcc-4.2` on 10.5/10.6 to try verify whether it fails or not.
A clear note is desirable, since otherwise it remains unclear why the port builds with gcc-4.2 but not apple-gcc-4.2. I don’t care if `apple-gcc-4.2` is blacklisted or not, but it must be clear that this does not apply to gcc-4.2 (at least in a sense the error was not observed with it so far; expanding backlist if needed is easy).

alex-free: Ah I see. For now let's just do the darwin8 conditional I'll get this changed.

barracuda156: A general note: it will be nice to have commits standardized in a way it is immediately clear what is changed.
```
brotli: import from MacPorts
brotli: blacklist Xcode gcc on Tiger
```
(I mean, start with a port name, then add a comment on what is done. Free form, but starting with a port name.)

alex-free: sure thing

alex-free:
@barracuda156 should be good

barracuda156:
@alex-free Please rebase to main (just drop commit importing the port), then we merge.

PR #196, opened 2026-08-05
gawk: remove upstreamed patches

alex-free:
Since Tiger patches were upstreamed, this is now failing to patch it. The reinpalce also is not needed anymore (no match).


alex-free:
MacPorts doesn't have the patch either: https://github.com/macports/macports-ports/blob/master/lang/gawk/Portfile, they still have the re-inplace, but thats harmless when it doesn't match

barracuda156 (approved):
Merged a bit differently, but from what it looks, it should work.

PR #197, opened 2026-08-06
libarchive: fix tiger build missing O_SYMLINK + fix compile when using gcc16

alex-free:
Also fixes other platforms using gcc-16


alex-free:
Issue is HAVE_COPYFILE_H expects O_SYMLINK to exist, which is an incorrect assumption on Tiger.

barracuda156:
> Also fixes other platforms using gcc-16

It certainly does not need any fixes on 10.6 with gcc16.

archivers/libarchive/Portfile
```diff
@@ -41,13 +41,17 @@ depends_lib-append \
...
 platform darwin 8 {
-    patchfiles-append   patch-libarchive-3.5-fix-tests-tiger.diff
+    patchfiles-append   patch-libarchive-3.5-fix-tests-tiger.diff patch-libarchive-write_disk_posix.c-tiger.diff
 }
 
 # 74145
 compiler.blacklist-append \
                 *gcc-4.0 *gcc-4.2 {clang < 421}
 
+if {[string match "macports-gcc-16" ${configure.compiler}]} {
```
barracuda156: No, we do not want unnecessary disabling of errors. (It is always easy to add it if really needed, gcc directly suggests what fails, but this is not the case where it is needed at all.)

alex-free:
> > Also fixes other platforms using gcc-16
> 
> 
> 
> It certainly does not need any fixes on 10.6 with gcc16.

If those flags are not needed for stricter gcc16 defaults on other then Darwin 8 I can wrap it in a conditional.

archivers/libarchive/files/patch-libarchive-write_disk_posix.c-tiger.diff
```diff
@@ -0,0 +1,19 @@
...
+ #if __APPLE__
+ #include <TargetConditionals.h>
++#include <AvailabilityMacros.h>
+ #if TARGET_OS_MAC && !TARGET_OS_EMBEDDED && HAVE_QUARANTINE_H
+ #include <quarantine.h>
+ #define HAVE_QUARANTINE 1
+ #endif
++/* Mac OS X 10.4 does not have O_SYMLINK, but when HAVE_COPYFILE_H is defined it is 
++assumed it will have it. */
++#if MAC_OS_X_VERSION_MIN_REQUIRED == MAC_OS_X_VERSION_10_4
```
barracuda156: 1. Numerical macros is preferred convention (`1040` over `MAC_OS_X_VERSION_10_4`).
2. Unless you mean that 10.3 does not need this fix, use a threshold, not an equality. It is possible someone compiles on a later OS but targeting 10.3.

barracuda156 (changes requested):

ForestExpertise:
> > > Also fixes other platforms using gcc-16
> > 
> > 
> > It certainly does not need any fixes on 10.6 with gcc16.
> 
> If those flags are not needed for stricter gcc16 defaults on other then Darwin 8 I can wrap it in a conditional.

-Wno-error=incompatible-pointer-types is the only one that should be needed, and only on Darwin 8.
May I ask what prompted the -Wno-error=implicit-function-declaration? I didn't need that and it is more problematic. If it was an acl function causing that error, I would recommend --disable-acl instead which I used to build on Tiger server.
I'll go ahead and close my PR for now.

archivers/libarchive/Portfile
alex-free: > No, we do not want unnecessary disabling of errors. (It is always easy to add it if really needed, gcc directly suggests what fails, but this is not the case where it is needed at all.)

These errors prevent the build from completing, so it is necessary to disable them. Gcc 16 suggested to do these. Again I can do it in a conditional, but I expected gcc to be gcc universally

alex-free: Perhaps the code path without COPYFILE_H is only affected by these errors so I'll just do the conditional since you've confirmed its not an issue on newer platforms

barracuda156: > Again I can do it in a conditional, but I expected gcc to be gcc universally

Please do it conditionally. These are _not_ needed because of gcc-16, they may be needed specifically on 10.4.
However, `-Wno-error=implicit-function-declaration`, if needed on your build, suggests some bug (which should be fixed instead of being ignored). `-Wno-error=incompatible-pointer-types` is fine to add, for 10.4 case.

archivers/libarchive/files/patch-libarchive-write_disk_posix.c-tiger.diff
alex-free: 10.3 doesn't have ACL at all, so it will never encounter this code as there would be many other errors before this.

archivers/libarchive/Portfile
alex-free: Done. Tiger ACL does not support symlinks, which was causing the error of implicit function declaration. Resolved this issue (simply doesn't set ACL on symlinks which is exactly what other software made for tiger does, see https://www.shoup.net/xbup/doc.pdf and search for 'acl_set_link_np'). Leopard has a way to do this by passing O_SYMLINK into acl_set_fd_np, which can easily be done with another define, but Tiger doesn't have O_SYMLINK :). I don't have Leopard right now so not writing support for something I can't verify.

alex-free:
closing and reopening

PR #202, opened 2026-08-08
rustdesk-ppc-agent: new port, and the minicargo fix it needs

danifunker:
A RustDesk agent -- the controlled side -- for PowerPC Macs: screen, keyboard, mouse and clipboard served to a standard RustDesk client, over a direct connection or a rendezvous server. There is no GUI client here. The protocol is reimplemented against RustDesk's own .proto files rather than ported from upstream's tokio-based tree, which is what keeps the graph at 32 crates and inside the subset of Rust mrustc handles.

Built and run with mrustc at d0fffe5b plus this tree's 0001-0025 and the 0026 added here, against rustc 1.90 libstd with debug assertions on. The whole crate graph builds -- protobuf, sodiumoxide, sha2, libc, five C shims and 36 MB of generated C -- and the resulting binary serves a full display probe on a Power Mac G5 and G4: framebuffer capture, ARGB conversion, VP8 encode, clipboard.

One crate needs patching. protobuf 3.0.0-alpha.2's BufReadIter reserves capacity and then indexes past the end of a zero-length slice, on a length taken mrustc needs 0026: minicargo parses a semver `+build` suffix but not `-prerelease`, so CARGO_PKG_VERSION reaches protobuf's build script as plain 3.0.0 -- it emits VERSION_3_0_0 while the crate's own generated code references VERSION_3_0_0_ALPHA_2, and the name fails to resolve. Submitted upstream as well; it applies above the CARGO_PKG_* block 0012 adds. revision bumped accordingly.

Also confirms 0002 subsumes the niche-enum alignment fix I had been carrying locally: protobuf's ReflectValueRef, the type that needed it, now compiles with no sizeof_assert failures.

### DIFF
--- a/lang/mrustc/Portfile
+++ b/lang/mrustc/Portfile
@@ -19,7 +19,7 @@ set rust_version_major_underscore [join [lrange [split ${rust_version} .-] 0 1] 
 
 # Subport mrustc-rust has its own versioning
 version             ${rust_version_major}-20260802
-revision            0
+revision            1
 epoch               1
 
 categories          lang devel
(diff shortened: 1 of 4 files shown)


barracuda156:
@danifunker A bit of a side question: can’t the build use a normal release of that crate instead of an alpha version of it?

danifunker:
Let me update this PR, I added more features to it so it's fully functional

PR #203, opened 2026-08-08
MPlayer: restore G4 and G5 variants

ForestExpertise:
+G4 greatly improves performance. I have not tested +G5

### DIFF
--- a/multimedia/MPlayer/Portfile
+++ b/multimedia/MPlayer/Portfile
@@ -75,3 +75,23 @@ x264 xmga xmms xss xv xvid xvmc xvr100 zr
 } {
     configure.args-append --disable-${option}
 }
+
+platform darwin powerpc {
+    variant G4 conflicts G5 description "Optimize for G4" {
+        configure.cflags-append \
+                    -mtune=G4 -mcpu=G4
+    }
+
+    variant G5 conflicts G4 description "Optimize for G5" {
+        # Only allow -mcpu=970 with old gcc 4.x. G5-specific codegen
+        # has bitrotten and does not work in the modern gcc:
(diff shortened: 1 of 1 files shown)


barracuda156 (changes requested):
Could you please pick flags from `ffmpeg`? I.e. use `-mcpu=970`, but disable powerpc64 when on ppc32.

ForestExpertise:
> Could you please pick flags from `ffmpeg`? I.e. use `-mcpu=970`, but disable powerpc64 when on ppc32.

Sure, I can match the way ffmpeg does it. I pulled from mplayer-devel, so perhaps the way mplayer-devel handles cpu variants should also be changed to match ffmpeg.

barracuda156:
@ForestExpertise Yeah, this is not a bug, just we updated it for `ffmpeg`, since the fix for codegen was found.

barracuda156:
@ForestExpertise Ok, one question: is `mplayer` C-only, or is there any C++ code? (ffmpeg has C++ in one plug-in of no importance, so I didn’t bother setting cxxflags too, but if mplayer has meaningful chunk of C++, then we want optflags in both C and C++.)

barracuda156:
Actually I can check that myself later. Gonna merge this now.

ForestExpertise:
> @ForestExpertise Ok, one question: is `mplayer` C-only, or is there any C++ code? (ffmpeg has C++ in one plug-in of no importance, so I didn’t bother setting cxxflags too, but if mplayer has meaningful chunk of C++, then we want optflags in both C and C++.)

I actually disabled the apple remote because it brought in c++ code from a system header which caused an error. Otherwise I cannot think of any c++ directly in mplayer.

PR #205, opened 2026-08-08
MPlayer: enable SDL support, revbump

ForestExpertise:
If there was a good reason to disable SDL support by default, I am open to doing this as a variant or making it Tiger specific. However, without SDL support, Tiger has no -vo option that really works. And I have a hard time seeing why the option of -vo SDL (and -vo gl if libsdl-cocoa is built with +opengl) is a bad thing for all platforms to have.


barracuda156 (approved):
I don’t know why it was off and why we want it on anyway, but w/e :)

PR #209, opened 2026-08-11
P5 xml parser: remove upstreamed tiger patch

alex-free:
No longer needed, patch fails to apply. changes implemented


barracuda156:
@alex-free Could you submit this also either as a PR to upstream MacPorts or as a ticket on Trac? 10.4 is not supported, so they should happily remove the patch, and then we can drop the port from here too.

PR #211, opened 2026-08-11
python314: don't use avx on tiger i386

alex-free:


lang/python314/Portfile
```diff
@@ -110,6 +110,13 @@ test.target         test
 
 destroot.target     frameworkinstall maninstall
 
+# Darwin 8 assembler can't generate these instructions, but gcc16 will try 
+# Unless all of these are set.
+if {${os.platform} eq "darwin" && ${os.major} == 8 && ${configure.build_arch} eq "i386"} {
```
barracuda156: I suggest to gate this on `${os.major} < 11` or even just i386 alone. (10.6 i386 with gcc cannot do these SIMD either, for example.)

barracuda156 (changes requested):

lang/python314/Portfile
barracuda156: P. S. There is no proper configure arg to turn these off?

alex-free: Surprisingly no. Maybe if you disable all optimizations (which is why Power is unaffected, probably), but thats such a waste. From what I've found it is supposed to be detect if these can be used. Newer GCC is 'aware' of these instructions, so it tries to use them when assembly optimizations are on, but the issue actually is when the assembler tries to build these instructions it has no idea what they are obviously. That's just saving further headache because if the assembler did know what these are, the built executable wouldn't run. Basically the oldest Intel platform is SSE3 and below optimizations as a baseline (GCC doesn't even have a specifc march for core duo, they tell you to do pentium 4 essentially). So yea let me know what you want the cutoff to be. Even with specifying march it STILL tries to use them so that was fun. IDK how many people are using gcc on intel as clang is probably more used, and that would probably handle it differently, but this wouldnt affect that in any way

barracuda156: Ok, just change the condition to `if {${os.platform} eq "darwin" && ${os.major} < 11 && ${configure.build_arch} eq "i386"} {` please (or plain i386 also fine).

barracuda156 (approved):

lang/python314/Portfile
```diff
@@ -110,6 +110,12 @@ test.target         test
 
 destroot.target     frameworkinstall maninstall
 
+# Python expects every x86 CPU to be able to use AVX/AVX2 with modern gcc. 
+if {${configure.build_arch} eq "i386"} {
+    configure.cflags-append "-D__AVX__=0 -D__AVX2__=0"
```
barracuda156: Sorry, last thing: do we need cflags at all? cppflags should already be passed into cflags.

alex-free: I will double check and rebuild.

PR #213, opened 2026-08-14
Orc: tiger specific and i386 fixes

alex-free:
fixes issues with older intel CPUs, and tiger. This is actually broken on anything remotely old in mainstream macports due to a variaty of reasons. One of them is the instruction that checks if things like AVX is available, is itself not found on older CPUs and there is no fallback.

### DIFF
--- a/devel/orc/Portfile
+++ b/devel/orc/Portfile
@@ -1,0 +1,42 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           meson 1.0
+PortGroup           muniversal 1.1
+
+name                orc
+version             0.4.42
+revision            0
+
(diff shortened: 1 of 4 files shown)


devel/orc/Portfile
```diff
@@ -39,4 +39,16 @@ platform darwin 8 {
...
+
+    if {${configure.build_arch} eq "i386"} {
+        patchfiles-append \
+            orc_orcprogram-x86.c_tiger.diff
+    }
+}
+
+if {${configure.build_arch} eq "i386"} {
+    patchfiles-append \
+        orc_orccpu-x86.c_i386.diff
```
barracuda156, line 53: Is it always correct for i386? Was it just broken before?

alex-free: yes always. No 32 bit Intel CPU will have the xgetbv instruction. The developers just never wrote or maintained a fallback for all the CPUs that don't have it (anything pre-2011)

alex-free: xgetbv is used to check for other advanced features like avx. But if you dont check if xgetbv itself is available, you run into this issue.

PR #208, opened 2026-08-08
brotli: blacklist apple gcc

alex-free:
All suggested changes implemented.

PR #214, opened 2026-08-14
libjpeg-turbo: blacklist apple gcc on intel

alex-free:
Nasm needs a new gcc, gcc16 works so we don't need to do the same thing as power does (below this) and can keep asm optimizations.

### DIFF
--- a/graphics/libjpeg-turbo/Portfile
+++ b/graphics/libjpeg-turbo/Portfile
@@ -47,6 +47,9 @@ if {${universal_possible} && [variant_isset universal]} {
 } elseif {${configure.build_arch} in {i386 x86_64}} {
     depends_build-append port:nasm
     configure.env       ASM_NASM=${prefix}/bin/nasm
+    # Need a newer gcc (such as gcc16) to provide this. Otherwise:
+    # error: immintrin.h: No such file or directory.
+    compiler.blacklist apple-gcc-4.2 *gcc-4.0
 } elseif {${configure.build_arch} eq "ppc"} {
     if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
         configure.args-append   -DWITH_SIMD=OFF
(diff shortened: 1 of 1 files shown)


graphics/libjpeg-turbo/Portfile
```diff
@@ -47,6 +47,9 @@ if {${universal_possible} && [variant_isset universal]} {
 } elseif {${configure.build_arch} in {i386 x86_64}} {
     depends_build-append port:nasm
     configure.env       ASM_NASM=${prefix}/bin/nasm
+    # Need a newer gcc (such as gcc16) to provide this. Otherwise:
+    # error: immintrin.h: No such file or directory.
+    compiler.blacklist apple-gcc-4.2 *gcc-4.0
```
barracuda156: This will pick gcc-4.2 which does not have `immintrin.h` either.

barracuda156 (changes requested):

barracuda156:
> so we don't need to do the same thing as power does (below this)

That code disables asm only if cpu does not have altivec (i.e. on G3).

graphics/libjpeg-turbo/Portfile
alex-free: this is the same as https://github.com/macos-powerpc/powerpc-ports/pull/208/changes/414a40abfaa5ecc44b8f9c9946b2ac101c3cf7e0 it blacklists both

alex-free:
> > so we don't need to do the same thing as power does (below this)
> 
> That code disables asm only if cpu does not have altivec (i.e. on G3).

Thats what I meant, newer gcc allows us to use the asm optimizations

graphics/libjpeg-turbo/Portfile
alex-free: oh, i see what u mean

alex-free: what do you suggest?

alex-free: ```*gcc-4.* ? or maybe *gcc-4.2 *gcc-4.0``` more conservatively?

barracuda156: Not something I care, and I don’t know if gcc 4.4–4.9 can handle this. None of those are picked by default anyway. Either is fine.

barracuda156 (approved):

PR #215, opened 2026-08-15
libdeflate: add more assembler checks for i386

alex-free:
Libdeflate already does this, but it falls short because it still is expecting a semi-modern Intel 64 bit CPU: https://github.com/ebiggers/libdeflate/issues/433

### DIFF
--- a/archivers/libdeflate/Portfile
+++ b/archivers/libdeflate/Portfile
@@ -34,6 +34,15 @@ github.tarball_from releases
 # https://github.com/ebiggers/libdeflate/commit/a83fb24e1cb0ec6b6fd53446c941013edf055192
 compiler.blacklist-append   *gcc-4.* {clang < 800} {macports-clang-3.[0-8]}
 
+# Libdeflate is aware that older binutils (like those in Xcode v2.5) do not 
+# support modern x86_64 instructions such as AVX512VNNI and AVX_VNNI and will
+# not use them if it detects the as will error out. This is expanded
+# to more instructions (AVX2, PCLMULQDQ, BMI2) for i386 to work.
+# See https://github.com/ebiggers/libdeflate/issues/433.
+if {${configure.build_arch} eq "i386"} {
+    patchfiles-append    additional_binutils_checks_i386.diff
+}
+
 configure.args-append \
                     -DLIBDEFLATE_BUILD_TESTS=1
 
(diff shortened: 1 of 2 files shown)


archivers/libdeflate/files/cmakelists.txt_i386.diff
```diff
@@ -0,0 +1,21 @@
+--- CMakeLists.txt.orig	2026-08-15 00:07:35.000000000 -0600
++++ CMakeLists.txt	2026-08-15 00:07:20.000000000 -0600
+@@ -155,6 +155,18 @@
+             # Set LIBDEFLATE_ASSEMBLER_DOES_NOT_SUPPORT_AVX_VNNI if needed.
+             check_assembler_support(AVX_VNNI "{vex} vpdpbusd %ymm0, %ymm0, %ymm0")
+         endif()
++        if(${CMAKE_C_COMPILER_VERSION} VERSION_GREATER_EQUAL 4.9)
```
barracuda156: Is there some reason to need 3 identical conditions instead of one?

archivers/libdeflate/Portfile
```diff
@@ -34,6 +34,19 @@ github.tarball_from releases
...
+# support modern x86_64 instructions such as AVX512VNNI and AVX_VNNI and will
+# not use them if it detects the as will error out. This is expanded
+# to more instructions (AVX2, PCLMULQDQ, BMI2) for i386 to work.
+# See https://github.com/ebiggers/libdeflate/issues/433.
+if {${configure.build_arch} eq "i386"} {
+    patchfiles-append \
+        cmakelists.txt_i386.diff \
+        lib_x86_adler32_impl.h_i386.diff \
+        lib_x86_crc32_impl.h_i386.diff \
+        lib_x86_decompress_impl.h_i386.diff
```
barracuda156: Let’s have 1 patch or, if you prefer, 2 (cmakelists + all headers stuff).

barracuda156 (changes requested):

archivers/libdeflate/files/cmakelists.txt_i386.diff
alex-free: I was just doing 1 for each to try to mirror what upstream is doing. I can have all fall under if gcc 4.9 or newer but the actual instruction checks should probably stay individual.

barracuda156: It looks like it should be three checks inside one `if(${CMAKE_C_COMPILER_VERSION} VERSION_GREATER_EQUAL 4.9)` macro.

barracuda156:
@alex-free Could you make the changes, so that we can merge this?

alex-free:
> @alex-free Could you make the changes, so that we can merge this?

Done

barracuda156 (approved):
I won’t test anything for i386, merging as is.

PR #216, opened 2026-08-15
bash: fix tiger

alex-free:


shells/bash/Portfile
```diff
@@ -123,6 +123,15 @@ variant universal {}
 
 compiler.blacklist-append *gcc-4.* {clang < 400}
 
+# ./trap.def:260:56: error: passing argument 2 of 'set_signal_handler' from
+# incompatible pointer type [-Wincompatible-pointer-types]
+if {${os.platform} eq "darwin" && ${os.major} == 8} {
```
barracuda156: Either `platform darwin 8` (easier to read) or `if {${os.platform} eq "darwin" && ${os.major} < 9} {`.

shells/bash/Portfile
```diff
@@ -135,6 +144,15 @@ configure.cflags-append -DSSH_SOURCE_BASHRC
 
 configure.checks.implicit_function_declaration.whitelist-append strchr
 
+# Because config.h is generated by configure, can't make a patch.
+# getenv.c:187:1: error: conflicting types for 'unsetenv'
+post-configure {
```
barracuda156: Better to consolidate both changes into one 10.4 block.

barracuda156 (changes requested):
I’d prefer to have

alex-free:
One platform darwin 8 is much better indeed

barracuda156:
Can you make this in 2 commits and not 3? I cannot squash this because we need import to be separate commit.

alex-free:
> Can you make this in 2 commits and not 3? I cannot squash this because we need import to be separate commit.

Done.

shells/bash/Portfile
```diff
@@ -123,6 +123,23 @@ variant universal {}
...
 
+platform darwin 8 {
+# ./trap.def:260:56: error: passing argument 2 of 'set_signal_handler' from
+# incompatible pointer type [-Wincompatible-pointer-types]
+    if {[string match "macports-gcc-16" ${configure.compiler}]} {
+        configure.cflags-append \
+            -Wno-error=incompatible-pointer-types \
+    }
+
+
```
barracuda156: No double empty lines, please.

shells/bash/Portfile
```diff
@@ -123,6 +123,23 @@ variant universal {}
 
 compiler.blacklist-append *gcc-4.* {clang < 400}
 
+platform darwin 8 {
+# ./trap.def:260:56: error: passing argument 2 of 'set_signal_handler' from
+# incompatible pointer type [-Wincompatible-pointer-types]
+    if {[string match "macports-gcc-16" ${configure.compiler}]} {
+        configure.cflags-append \
+            -Wno-error=incompatible-pointer-types \
```
barracuda156: Backslash should not be here, there is no line break.

shells/bash/Portfile
```diff
@@ -123,6 +123,23 @@ variant universal {}
 
 compiler.blacklist-append *gcc-4.* {clang < 400}
 
+platform darwin 8 {
+# ./trap.def:260:56: error: passing argument 2 of 'set_signal_handler' from
+# incompatible pointer type [-Wincompatible-pointer-types]
+    if {[string match "macports-gcc-16" ${configure.compiler}]} {
```
barracuda156: Since this anyway needs to be fixed, better also change this line to `if {[string match "macports-gcc*" ${configure.compiler}]} {` so that this survives gcc-17 later.

barracuda156 (changes requested):

barracuda156 (approved):

PR #218, opened 2026-08-16
boehmgc: fix tiger build

alex-free:
Would love some feedback on why this even being passed in the first place

### DIFF
--- a/devel/boehmgc/Portfile
+++ b/devel/boehmgc/Portfile
@@ -1,0 +1,43 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           muniversal 1.0
+
+github.setup        ivmai bdwgc 8.2.12 v
+revision            0
+checksums           rmd160  dc1504fdb9da1808664a34f5f9ad38d983571b9c \
+                    sha256  42e5194ad06ab6ffb806c83eb99c03462b495d979cda782f3c72c08af833cd4e \
+                    size    1231996
+
+name                boehmgc
+categories          devel
+maintainers         nomaintainer
+license             Permissive
+
+description         The Boehm-Demers-Weiser conservative C/C++ Garbage Collector
+long_description \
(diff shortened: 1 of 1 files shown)


devel/boehmgc/Portfile
```diff
@@ -29,6 +29,11 @@ distname            gc-${version}
 # gc_atomic_ops.h: error: '__ATOMIC_RELAXED' undeclared (first use in this function)
 compiler.c_standard 2011
 
+# g++-mp-16: error: unrecognized command-line option '-stdlib=libstdc++' 
+if {${os.platform} eq "darwin" && ${os.major} == 8} { 
+    configure.cxx_stdlib
```
barracuda156, line 34: Why is being added at all though? This is probably not the way it should be fixed here.

alex-free: Right? I think it's something in the build system, I'm going to take a more thorough look I've just been trying to crunch through the broken dep chain for pulse audio a little too fast.

barracuda156: The outcome that will be correct:
1. Whenever `libstdsc++` is the system default, `-stdlib=libstdc++` is never added for gcc.
2. Whenever `libc++` is the system default, `-stdlib=libc++` is passed to gcc (clang should not need it).

Resetting `configure.cxx_stdlib` to nothing is a fragile hack that may work in a given environment but does everything in a wrong way.

alex-free: My thinking is boehmgc always expects clang so it is passing that, even though this is obviously not clang. I'll let you know what I find

barracuda156: There is nothing in portfile that would pull this: https://github.com/macports/macports-ports/blob/c3d1c85ba53e8e9f0a7b6032e02bbf1fa670c0fd/devel/boehmgc/Portfile

So unless you have some local tweak, it is either the base (i.e. bug is either mine or MacPorts’s) or upstream code itself.

We need to find what adds the flag, and fix that, instead of hacking around it.

barracuda156: The flag is not passed on my 10.6 (I do not have 10.6 i386, so it still could be arch-dependent).

alex-free:
@barracuda156 it was because of https://github.com/macos-powerpc/ppcports-base/blob/6c4a3c5b526f9d8f611ade19f937475244e3744c/src/port1.0/portconfigure.tcl#L87

Added a check for if snow leopard or below on Intel and it's fine

PR #220, opened 2026-08-16
atk: disable docs on tiger

alex-free:


devel/atk/Portfile
```diff
@@ -38,6 +38,13 @@ gobject_introspection   yes
 
 configure.args      -Ddocs=true
 
+# atk-scan is linked with @loader_path/libatk-1.0.0.dylib during doc
+generation phase, which is incorrect. Skipping docs prevents it from being
```
barracuda156: Looks like this will fail to parse.

alex-free: I'm going to redo this

alex-free:
Confirmed that -Ddocs=true is actually the default now (perhaps it was not at some point in the past). So now only on darwin 8 it needs to be specified as false

barracuda156:
Please do not change general defaults here. It is useful to have them explicit. What is needed is `configure.args-replace`.

alex-free:
> Please do not change general defaults here. It is useful to have them explicit. What is needed is `configure.args-replace`.

Done

barracuda156 (approved):

PR #221, opened 2026-08-17
mesa: fix i386 build

alex-free:


x11/mesa/Portfile
```diff
@@ -75,14 +75,34 @@ patchfiles-append       patch-gallium_vl_csc_c.diff
 # patchfiles-append       patch-build-with-LLVM-7.diff
 # Disable JIT on ppc, it is anyway broken in LLVM.
 # See also: https://gitlab.freedesktop.org/mesa/mesa/-/work_items/16103
-patchfiles-append       patch-no-jit-on-ppc.diff
+# Only do it if not darwin 8, because on darwin 8 we disable it for both ppc
+# and i386, because of the below error with gcc16:
+# os_misc.c:543:52: error: 'MAP_JIT' undeclared
+# It may be declared on darwin 9 (via llvm?) or newer, so don't disable it on 
+# all i386 platforms until confirmed when it is available.
+if {${os.platform} eq "darwin" && ${os.major} != 8} {
```
barracuda156: This is unnecessarily messy way to do it. Better options:
1. Inside 10.4 case use `patchfiles-replace` for this one.
2. Use a patch for 10.4 that applies on top of existing one.
Either way, leave general patch alone, let it apply unconditionally.

While the trigger was `MAP_JIT`, JIT will not work on powerpc regardless, that is a separate thing. `MAP_JIT` should be conditional on macOS version, I just did not bother with that, waiting for Mesa folks to fix a bug they introduced. See: https://gitlab.freedesktop.org/mesa/mesa/-/work_items/16103

barracuda156 (changes requested):

alex-free:
All suggestions applied.

x11/mesa/files/patch-no-jit-on-tiger.diff
```diff
@@ -0,0 +1,22 @@
+--- src/util/os_misc.c
++++ src/util/os_misc.c  2026-08-16 11:34:56.000000000 +0800
+@@ -520,7 +521,7 @@
+ #endif
+ }
+
+-#if DETECT_OS_APPLE
++#if DETECT_OS_APPLE && !defined(__POWERPC__) && !defined(__i386)
```
barracuda156: Wrong macro here for i386

x11/mesa/Portfile
```diff
@@ -77,12 +77,30 @@ patchfiles-append       patch-gallium_vl_csc_c.diff
...
 # restrict to Tiger for now, therefore
+# On Tiger  we disable jit for ppc and i386, because MAP_JIT is not available:
+# os_misc.c:543:52: error: 'MAP_JIT' undeclared
 platform darwin 8 {
-    patchfiles-append   patch-mesa-apple-cgi-int-differences.diff
+    patchfiles-append \
+        patch-mesa-apple-cgi-int-differences.diff
+    
+    patchfiles-replace \
+        patch-no-jit-on-ppc.diff \
```
barracuda156: This patch no longer exists, this won’t work at all.

barracuda156 (changes requested):

x11/mesa/Portfile
alex-free: The tree is out of sync at this point. Does a new pull request make sense here, or is this fine with the recent force push?

alex-free:
Redone with updated sources in new pull request, closing

PR #222, opened 2026-08-17
pango: fix build on 10.4

alex-free:


x11/pango/Portfile
```diff
@@ -66,6 +66,13 @@ compiler.cxx_standard   2011
 configure.cxxflags-append \
                         -std=c++11
 
+# viewer-render.c:987:49: error: passing argument 3 of 'g_file_get_contents'
+# from incompatible pointer type
+if {[string match "macports-gcc-16" ${configure.compiler}]} {
```
barracuda156: This is not needed for gcc16. If 10.4 needs it, please add it for 10.4 (perhaps without gcc16 case, otherwise it breaks again with gcc17).

barracuda156 (changes requested):

x11/pango/Portfile
```diff
@@ -66,6 +66,15 @@ compiler.cxx_standard   2011
 configure.cxxflags-append \
                         -std=c++11
 
+# viewer-render.c:987:49: error: passing argument 3 of 'g_file_get_contents'
+# from incompatible pointer type
+platform darwin 8 {
+    if {[string match "macports-gcc-16" ${configure.compiler}]} {
```
barracuda156: Same note: use a general `macports-gcc*` form so that we do not need to change this every time gcc gets updated.

barracuda156 (changes requested):

barracuda156 (approved):

PR #225, opened 2026-08-18
fontconfig: use same xz tarball like MacPorts so that fetching from mirrors work

alex-free:
Please check that I didn't break gitlab pull. But I got the tarball directly from gitlab (they only offer xz now starting this release) so I had to change that. Also, no http distfile mirror has been refreshed yet to include the new xz, but it's working from [http://tigerports.com/macports/distfiles](http://tigerports.com/macports/distfiles) fine

### DIFF
--- a/graphics/fontconfig/Portfile
+++ b/graphics/fontconfig/Portfile
@@ -18,9 +18,14 @@ long_description            Fontconfig is a library for configuring \
                             and customizing font access.
 
 homepage                    https://www.freedesktop.org/wiki/Software/fontconfig
-checksums                   rmd160  0a84e9647538eb34f40fba33bb058426ec2a4256 \
-                            sha256  8d28f79d2017cbe1fbb7da84b2502c86421b4f45860234d2f4ab5b35564c8d01 \
-                            size    544066
+# the tarball from GitLab has no configure script, this appears to be the "release"
+# but the gitlab PG does not support this. See: https://trac.macports.org/ticket/71113
+master_sites                https://gitlab.freedesktop.org/api/v4/projects/890/packages/generic/fontconfig/${version}
+use_xz                      yes
+
+checksums                   rmd160  5a05f7ffd0d4794831e720f37cefc3ad11bebbfd \
+                            sha256  4f7b554a38cdf78c033f666c8871f3749e14a094f65a07f630c91ed0b43d35e3 \
+                            size    1172016
 
 set python_version          3.14
 


barracuda156:
Why, what’s up with this? It works fine as-is.

alex-free:
> Why, what’s up with this? It works fine as-is.

Original Mac OS X curl via http does not work and requests tar expecting a .tar.bz2, so completely broken without updated curl

barracuda156:
Well, then curl should be fixed, not the port. This is the set-up we are using on 10.5–10.6. Having a broken curl is not a livable option anyway.

alex-free:
> Well, then curl should be fixed, not the port. This is the set-up we are using on 10.5–10.6. Having a broken curl is not a livable option anyway.

So if the gitlab repo were to go down (as unlikely as that would be) we dont want to fallback to downloading a tarball via distfiles.macports.com or any other mirror site like any other port? That also not work with newer curl as the distfile it's trying to pull is entirely wrong

barracuda156:
> So if the gitlab repo were to go down (as unlikely as that would be) we dont want to fallback to downloading a tarball via distfiles.macports.com or any other mirror site like any other port? That also not work with newer curl as the distfile it's trying to pull is entirely wrong

You phrase it in a way as if something is currently broken, but it is not. Everything works. Which is why I was objecting, IMO quite naturally.
However it does make sense to use the same tarball which MacPorts uses, so that fetching from there will work too. I am fine with borrowing this from MacPorts: https://github.com/macports/macports-ports/blob/d93bcd3aae829be4766b871afd895115ab7a3ec3/graphics/fontconfig/Portfile#L23-L26
Will this address your concern?

alex-free:
> > So if the gitlab repo were to go down (as unlikely as that would be) we dont want to fallback to downloading a tarball via distfiles.macports.com or any other mirror site like any other port? That also not work with newer curl as the distfile it's trying to pull is entirely wrong
> 
> You phrase it in a way as if something is currently broken, but it is not. Everything works. Which is why I was objecting, IMO quite naturally. However it does make sense to use the same tarball which MacPorts uses, so that fetching from there will work too. I am fine with borrowing this from MacPorts: https://github.com/macports/macports-ports/blob/d93bcd3aae829be4766b871afd895115ab7a3ec3/graphics/fontconfig/Portfile#L23-L26 Will this address your concern?

Yea that would be great, it just needs those lines really. Appreciate it. I phrased the initial response poorly thats on me. I meant the entire url it falls back to is currently incorrect for mirror sites (because of file extension defaulting to .tar.bz2, and no master site listed) (and in turn in my fork my http website is a mirror site, but that is irrelevant)

barracuda156:
If you change the PR to just borrow those lines (and adjust checksums), this can be merged. (No need to modify extraction command, by the time someone builds fontconfig xz must exist.)

alex-free:
> If you change the PR to just borrow those lines (and adjust checksums), this can be merged. (No need to modify extraction command, by the time someone builds fontconfig xz must exist.)

Seems like the issue with why I need the extract command is because of a bug in base: https://trac.macports.org/ticket/71113 that was fixed after tiger_upd branch of ppcports-base was last at parity . That's why it always tries .tar.bz2. Things like this is why I need to get parity with the main ppcports-base branch.

Not relevant though for this pull request. Will do just the lines from macports and updated checksums

PR #226, opened 2026-08-18
py-gobject3: fix tiger build with gcc16

alex-free:


barracuda156 (changes requested):
Please align the added block correctly, otherwise looks fine.

PR #227, opened 2026-08-18
libsdl2-cocoa: enable audio on tiger

alex-free:
Tested some sample audio programs playing wavs and whatnot and no problems. Will be testing things like video players soon.

Must sudo port load everything for dbus and pulseaudio that it asks for after install to get this to work. Note: dbus has a system wide user agent that wont work on ppc-ports base for tiger specifically. (there is a workaround i did for tigerports-base and it's working fine there). I'll try to submit the workaround into ppcports-base tiger_upd branch if there is interest.

### DIFF
--- a/devel/libsdl2-cocoa/Portfile
+++ b/devel/libsdl2-cocoa/Portfile
@@ -127,10 +127,19 @@ platform darwin 10 {
                     +oldapi
 }
 
-# FIXME: look into fixing CoreAudio module for Tiger.
+# Coreaudio is completely different on 10.5 and newer. Use pulseaudio instead.
 if {${os.platform} eq "darwin" && ${os.major} < 9} {
+    depends_build-append port:pulseaudio
+
+    configure.args-replace \
+                    --disable-pulseaudio \
+                    --enable-pulseaudio
+
+    # Normally on Mac OS X configure forces coreaudio, and never even checks 
+    # for pulseaudio. This makes it only check for coreaudio.
+    patchfiles-append    configure_tiger.diff
+
     configure.args-append \
-                    --disable-audio \
                     --disable-haptic \
                     --disable-hidapi \
                     --disable-hidapi-joystick \
--- a/devel/libsdl2-cocoa/Portfile
+++ b/devel/libsdl2-cocoa/Portfile
@@ -128,15 +128,17 @@ platform darwin 10 {
 }
 
 # Coreaudio is completely different on 10.5 and newer. Use pulseaudio instead.
-if {${os.platform} eq "darwin" && ${os.major} < 9} {
-    depends_build-append port:pulseaudio
+platform darwin 8 {
+    # Revision bump for audio support on tiger.
+    incr revision
+    depends_lib-append port:pulseaudio
 
     configure.args-replace \
                     --disable-pulseaudio \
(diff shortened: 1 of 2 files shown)


alex-free:
@ForestExpertise https://github.com/macos-powerpc/ppcports-base/pull/4

ForestExpertise:
> @ForestExpertise [macos-powerpc/ppcports-base#4](https://github.com/macos-powerpc/ppcports-base/pull/4)

Quick note: this should include a rev bump to force rebuilds. I am unsure if I can take a look at this right now on hardware, because it is over a hundred degrees Farenheit this week and I only have a wall unit for air conditioning. But it looks like great work!

alex-free:
> > @ForestExpertise [macos-powerpc/ppcports-base#4](https://github.com/macos-powerpc/ppcports-base/pull/4)
> 
> Quick note: this should include a rev bump to force rebuilds. I am unsure if I can take a look at this right now on hardware, because it is over a hundred degrees Farenheit this week and I only have a wall unit for air conditioning. But it looks like great work!

Good idea. I want to update a comment anyways

alex-free:
Now has a revision for darwin 8, should be good to go

devel/libsdl2-cocoa/Portfile
```diff
@@ -127,10 +133,19 @@ platform darwin 10 {
                     +oldapi
 }
 
-# FIXME: look into fixing CoreAudio module for Tiger.
+# Coreaudio is completely different on 10.5 and newer. Use pulseaudio instead.
 if {${os.platform} eq "darwin" && ${os.major} < 9} {
+    depends_build-append port:pulseaudio
```
barracuda156: Is this only build-time dependency??

devel/libsdl2-cocoa/Portfile
```diff
@@ -8,6 +8,12 @@ name                libsdl2-cocoa
 conflicts           libsdl2-powerpc libsdl2-snowleopard libsdl2-x11
 github.setup        libsdl-org SDL 2.32.10 release-
 revision            2
+
+# Enable audio with pulseaudio.
+platform darwin 8 {
+    revision            3
```
barracuda156: Please remove this from here and place inside 10.4 block below.
It is also easier to do `incr revision` instead of hardcoding the number.

barracuda156 (changes requested):

devel/libsdl2-cocoa/Portfile
alex-free: I'm not sure what you mean by this. There is a .pc file that could go here instead (I don't know what is more correct and I've seen that on other portfiles), but this is the only additional dependency needed on darwin 8 versus other platforms.

barracuda156: `depends_build` means that it is not linked to and not used by runtime. So it will not be installed/activated, if libsdl2 is already built.

alex-free: I see

alex-free:
Panther support would require additional patches at the very least (I ported SDL2.0.3 to panther so I know), so just put everything in platform darwin 8 block.

barracuda156 (approved):

PR #229, opened 2026-08-18
dbus: fix linking on tiger

alex-free:
I _really dont like this _hack__, but this is something that was needed to get libsdl2-cocoa working with pulseaudio as it's a pulseaudio dependency. I wanted to wait until I figured out a better fix but now that libsdl2-cocoa+pulseaudio is working on tiger I want to at least post this so that the entire dependency chain fixes can be known.  Perhaps @ForestExpertise could take a look too. This is the last fix I hadn't yet submitted for the pulse audio tiger dep chain.

### DIFF
--- a/devel/dbus/Portfile
+++ b/devel/dbus/Portfile
@@ -109,6 +109,18 @@ pre-destroot {
 }
 
 post-destroot {
+    # On Tiger dbus-daemon-launch-helper is trying to load the dylib from the
+    # wrong folder because the dylib it's linked with has @loader_path but the
+    # dylib is not in the same folder as dbus-daemon-launch-helper. It is
+    # installed to the correct place (verified on supported mac os with
+    # official macports) so for now just fix it with install_name_tool.
+    if {$macosx_deployment_target eq "10.4"} {
+        system "install_name_tool -change \
+            '@loader_path/libdbus-1.3.dylib' \
+            '@loader_path/../lib/libdbus-1.3.dylib' \
+            ${destroot}${prefix}/libexec/dbus-daemon-launch-helper"
+    }
+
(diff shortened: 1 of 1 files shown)


ForestExpertise:
> I _really dont like this _hack__, but this is something that was needed to get libsdl2-cocoa working with pulseaudio as it's a pulseaudio dependency. I wanted to wait until I figured out a better fix but now that libsdl2-cocoa+pulseaudio is working on tiger I want to at least post this so that the entire dependency chain fixes can be known. Perhaps @ForestExpertise could take a look too. This is the last fix I hadn't yet submitted for the pulse audio tiger dep chain.

Looks like great work, Alex! I had just been suffering through dbus having a broken binary, this is much better. May I suggest putting it in a darwin 8 block? The post-destroot code should still work fine there.

barracuda156:
I agree, this should be inside 10.4-specific block. (Whether retaining deployment target condition or not is up to you.)

ForestExpertise:
Oh, and we should ideally rev_bump inside the Tiger block for anyone who has this built already, as this is not a build failure

alex-free:
Ok rev bump applied, platform darwin 8 used in destroot, rebuilt and tested working

devel/dbus/Portfile
```diff
@@ -6,7 +6,7 @@ PortGroup       muniversal 1.0
 
 name            dbus
 version         1.16.2
-revision        0
+revision        1
```
barracuda156: If 10.4 needs a revbump, please keep it to 10.4. Also that trailing space.

barracuda156 (changes requested):

PR #231, opened 2026-08-19
wayland: fix OSX 10.9 build with default MacPorts

nilFinx:
* pins c verison with meson to c11 for stdatomic.h
* blacklist apple clang 6 which lacks stdatomic.h afaik

alternative approach is to disable tests, but yeah, not a good idea.
test passes for me... except for two. display and compositor-introspection. irrelevant for this patch because stdatomic.h not found was in compositor test.

### DIFF
--- a/wayland/wayland/Portfile
+++ b/wayland/wayland/Portfile
@@ -79,12 +79,20 @@ if {${subport} eq "${name}"} {
     # stdatomic.h
     compiler.c_standard 2011
 
+    # because Apple Clang 6 on OS X 10.9 appears to lack stdatomic.h
+    compiler.blacklist-append {clang < 602}
+
     configure.args-append \
                     -Ddocumentation=false \
                     -Ddtd_validation=true \
                     -Dlibraries=true \
                     -Dscanner=true \
                     -Dtests=true
+
+    if [string match *clang* ${configure.compiler}] {
+        configure.args-append \
+                    -Dc_std=c11
+    }
 
     variant docs description "Install docs" {
         configure.args-replace \


wayland/wayland/Portfile
```diff
@@ -79,12 +79,16 @@ if {${subport} eq "${name}"} {
...
+    compiler.blacklist-append {clang < 602}
+
     configure.args-append \
                     -Ddocumentation=false \
                     -Ddtd_validation=true \
                     -Dlibraries=true \
                     -Dscanner=true \
-                    -Dtests=true
+                    -Dtests=true \
+                    -Dc_std=c11
```
barracuda156, line 94: @nilFinx Thank you. Could you please restrict `-Dc_std=c11` to where it is actually needed?

barracuda156 (changes requested):

wayland/wayland/Portfile
nilFinx: what do you mean?

barracuda156: Forcing C11 will produce different code, so it is undesirable unless required (and in general this is not required; I cannot test on 10.9 and trust you on that).

nilFinx: Hmmm.... Then, how about making only the tests compile as C11, if that's possible?

barracuda156:
@nilFinx BTW did you get the whole wayland stack working? `foot`, `swayimg`, `mpv`

nilFinx:
Not even Clang 17 works on foot and swayimg thanks to MacPorts' epoll-shim causing forward declarations.. I can send a commit for bumping legacysupport to 10.11 (for CLOCK_MONOTONIC?) though.

barracuda156:
swayimg needs C++23 and fails on 10.15, for example (I forgot about that), but foot should work. I will verify the build. Make sure to use ports from here in the latest versions.

If LegacySupport is already used, then yes, please adjust version threshold as needed. I would not catch clock_gettime, if a build failed earlier on strnlen, for example.

P. S. I will update the whole stack on 10.15 now and fix w/e fails, if anything. Once that works, your 10.9 is feasible automatically: it just sits in between 10.6 (known to work) and 10.15 (was working, to be re-confirmed).

barracuda156:
@nilFinx 
wl

`swaying` aside, up to `foot` everything should work.

barracuda156:
@nilFinx This should get you image viewer w/out C++23: https://github.com/macos-powerpc/powerpc-ports/commit/74a740cc0c1c3ef97bcecdb39f53ea49d1d180f4
Tested on 10.15, works both in wayland and x11.

barracuda156:
I got SDL2 working over wayland now, including video support. Gonna push soon.

barracuda156:
@nilFinx Hopefully the merged version satisfies you, I just restricted C11 to clang, so that gcc is not affected.

I pushed all updates, now, at least provisionally, sdl2/sdl3, gtk3 and mpv should work over wayland.

PR #232, opened 2026-08-22
aqua/aqualink: add draft Portfile

watermark-hd:
Draft Portfile for AquaLink (SMB2/3 client + WebDAV NAS bridge for Tiger/Leopard PowerPC Macs), per the suggestion in the MacRumors thread: https://forums.macrumors.com/threads/aqualink-smb2-3-client-for-powerpc-tiger-leopard-mounts-modern-nas-shares-in-finder.2487262/

Upstream source: https://github.com/watermark-hd/ppc-mac-modernization (tag v0.2)

## Known gaps / things I'd appreciate a look at

- This vendors and builds libsmb2 6.0.0 inline (bootstrap/configure/make) rather than depending on a proper `libsmb2` port, since one doesn't exist in this tree yet. Happy to split it out into its own port if that's preferred.
- Build/destroot logic is written to match the mkconsole Portfile's general shape, but I don't have a local MacPorts install to `port lint` or test-build this against, so structural mistakes are likely.
- `license` is set to `unknown` — the upstream repo doesn't declare one explicitly yet, will fix upstream and update this.
- `maintainers` left as `nomaintainer` for now.

This is intentionally a rough first pass, not a finished PR — flagging as draft so it's clear it needs a look before merging.

### DIFF
--- a/aqua/aqualink/Portfile
+++ b/aqua/aqualink/Portfile
@@ -1,11 +1,13 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           makefile 1.0
 
 name                aqualink
-version             0.2
+github.setup        watermark-hd ppc-mac-modernization 0.3 v
+revision            0
 categories          aqua net
-platforms           darwin
 license             unknown
 maintainers         nomaintainer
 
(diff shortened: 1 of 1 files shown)


barracuda156:
@watermark-hd Thank you!

I have added `libsmb2` port now: https://github.com/macos-powerpc/powerpc-ports/commit/ff31393646cde1c8357369da6d514865519c2198

barracuda156:
@watermark-hd This works on my end:

Fix Makefile:
```
--- Makefile	2026-08-20 18:19:03.000000000 +0800
+++ Makefile	2026-08-22 14:22:00.000000000 +0800
@@ -3,13 +3,13 @@
 CONTENTS = $(APPBUNDLE)/Contents
 MACOS = $(CONTENTS)/MacOS
 
-LIBSMB2_DIR = $(HOME)/developer/libsmb2
+LIBSMB2_DIR ?= $(PREFIX)
 LIBSMB2_INC = $(LIBSMB2_DIR)/include
-LIBSMB2_A = $(LIBSMB2_DIR)/lib/.libs/libsmb2.a
+LIBSMB2_A = $(LIBSMB2_DIR)/lib/libsmb2.dylib
 
-CC = gcc
-CFLAGS = -std=gnu99 -Wall -fobjc-exceptions -I$(LIBSMB2_INC)
-LDFLAGS = -framework Cocoa -framework Security
+CC ?= gcc
+CFLAGS += -std=gnu99 -Wall -fobjc-exceptions -I$(LIBSMB2_INC)
+LDFLAGS += -framework Cocoa -framework Security
 
 OBJS = main.o AppDelegate.o WebDAVServer.o LocalWebDAVServer.o
 
```

Portfile:
```
# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4

PortSystem          1.0
PortGroup           github 1.0
PortGroup           makefile 1.0

name                aqualink
github.setup        watermark-hd ppc-mac-modernization 0.2 v
revision            0
categories          aqua net
# license             MIT
maintainers         nomaintainer

description         SMB2/3 client and WebDAV NAS bridge for PowerPC MacOS

long_description    AquaLink lets PowerPC Macs running Tiger or Leopard connect to \
                    modern SMB2/3 NAS shares, which the OS's built-in SMB1-only \
                    Samba stack cannot speak. It provides a Transmit/Cyberduck-style \
                    file browser for drag-and-drop transfers, and a local WebDAV \
                    bridge that lets the same share be mounted as a normal Finder \
                    volume via mount_webdav, without any kernel extension. It also \
                    includes the reverse: exposing a folder on the old Mac as \
                    a WebDAV share that a modern Mac or Windows machine can mount.

checksums           rmd160  497a9d7a3f24f014d9a43e83762d194b1e1f6818 \
                    sha256  ffc332ad4b29ebf9e5e7faede02a5077daced9c140b214726f2c613cbdb6c3df \
                    size    72515
github.tarball_from archive

worksrcdir          ppc-mac-modernization-${version}/smb3/AquaLink

patchfiles          patch-Makefile.diff

depends_lib-append  port:libsmb2

compiler.c_standard 1999

build.args          LIBSMB2_DIR=${prefix}

destroot {
    copy ${worksrcpath}/AquaLink.app ${destroot}${applications_dir}
}

livecheck.type      regex
livecheck.url       https://github.com/watermark-hd/ppc-mac-modernization/tags
livecheck.regex     {v(\d+(?:\.\d+)*)}
```

barracuda156:
@watermark-hd If you make a few changes, we will be able to merge this

watermark-hd:
Since libsmb2 has been ported, I’ve switched to making it depend on `port:libsmb2` via `depends_lib`, as you suggested (this aligns with the GitHub and Makefile PortGroup configuration).

I’ve arranged it so that no patch files are needed. The Makefile for AquaLink itself has been updated in the upstream version (v0.3) to automatically detect whether a dylib or static library is installed simply by passing a single `LIBSMB2_DIR` value, so adding `LIBSMB2_DIR=${prefix}` to `build.args` is all that’s needed.

I’ve updated the version to v0.3. This includes changes from v0.2 onward (an administrator-privileged unmount fallback, and a warning about plaintext HTTP communication).

watermark-hd:
Here’s a follow-up report. After adding the SMB3 encryption-required option to AquaLink, I encountered a bug where the connection would be disconnected immediately after successful authentication for both standard SMB shares on macOS and Windows 11.

After capturing and analyzing packets on a physical machine (a PowerPC/big-endian iBook), I found a bug in `lib/smb3-seal.c` of libsmb2 where the endian conversion was omitted only for the SessionId field in the SMB3 encryption header (while other fields are converted using htole32/htole16, only session_id is copied using a raw memcpy). It appears this bug went unnoticed because it happened to work on x86 and ARM platforms.

I’ve reported this to the libsmb2 author and opened an issue with a fix included: https://github.com/sahlberg/libsmb2/issues/477

If possible, please apply this patch to the PPCPorts devel/libsmb2 Portfile (until the official updated version is released). This will allow AquaLink’s SMB3 encryption requirement option to work via PPCPorts as well.

barracuda156:
@watermark-hd Would you like to make a PR for `libsmb2` patch here? Your port is the only consumer, and I can rely on results of your testing, so it can be merged right-away, then we follow up with this PR merging.

barracuda156 (approved):

barracuda156:
@watermark-hd Thank you!

PR #233, opened 2026-08-27
mesa: remove no longer applying Tiger patch

ForestExpertise:
The mesa patch no longer applies, and mesa seems to both build and work without it, so I think it is okay to drop.
Libepoxy wants +egl, so I see no harm in having that as a default variant for Tiger. @alex-free feel free to chime in, I am open to better proposed changes if they are worth the effort.


x11/mesa/Portfile
```diff
@@ -102,6 +94,8 @@ platform darwin {
     if {${os.major} < 9} {
           # Xplugin.h is missing on Tiger
           configure.cppflags-append -I${filespath}/include
+          #libepoxy wants this
+          default_variants +egl
```
barracuda156: This should go below where default variants are set. Yes, we generally want Tiger stuff to be in one block, but not in this case.

barracuda156 (changes requested):

x11/mesa/Portfile
ForestExpertise: It is now with the other blocks that set default variants

barracuda156:
First: +wayland pulls in +egl, i.e. non-darwin do not need the latter as a special case at all.

Secondly: epoxy does not require egl as such, if it is enabled by default, it is to support some wayland-related port. Since there is no wayland on 10.4, an alternative is not to require +egl in libepoxy for 10.4.

I am fine with requiring +egl in mesa for Tiger, it’s just not the only solution.

alex-free:
It looks like they removed apple_glx_diagnostic() completely so the patch can go.

Maybe libepoxy should be changed rather then the other way around, as not everything needs +egl variant.

barracuda156:
> as not everything needs +egl variant.

Nothing actually needs `+egl` aside of Wayland, AFAIK. I enabled those by default because it is just annoying for me to build `mesa` in 3 variants on every upgrade. If Wayland is disabled on 10.4, there is no reason to have EGL anywhere – nothing consumes it.

x11/mesa/Portfile
```diff
@@ -185,6 +177,11 @@ if {${os.platform} ne "darwin" || ${os.major} >= 10} {
     default_variants        +wayland
 }
 
+if {${os.platform} ne "darwin" || ${os.major} <= 8} {
```
barracuda156: Regardless of what is set for 10.4, please remove `${os.platform} ne "darwin"` part. There is nothing in common between Linux and Tiger case, they should not be grouped, even if incidentally they may use same variant (but Linux should rather use `+wayland`).

barracuda156 (changes requested):

alex-free:
This seems to want some slightly different dependencies down the line, like py314-cairo+xquartz+x11 (rather then just +x11) if you go with +egl here while building libsdl2 on tiger). Without +egl requirement on libepoxy it goes back to just needing py314-cairo+x11 after a port sync

ForestExpertise:
> This seems to want some slightly different dependencies down the line, like py314-cairo+xquartz+x11 (rather then just +x11) if you go with +egl here while building libsdl2 on tiger). Without +egl requirement on libepoxy it goes back to just needing py314-cairo+x11 after a port sync

Well, that's enough to convince me, especially since you are handling the libepoxy part. I simplified this down to removing the no longer applying patch. @barracuda156 we should be ready to merge now.

barracuda156 (approved):

PR #235, opened 2026-08-27
doxygen: use fallback on Tiger, fix for gcc16

ForestExpertise:
Since doxygen is almost always a build time dependency, it is an ideal candidate for a fallback. Since the current version doesn't build, this seems like the path of least resistance. @alex-free if you would prefer to fix the current build, I have no objection.

### DIFF
--- a/textproc/doxygen/Portfile
+++ b/textproc/doxygen/Portfile
@@ -105,19 +105,19 @@ if {${os.platform} eq "darwin" && ${os.major} < 13} {
                         path:bin/git:git
     configure.args-append \
                         -DGIT_EXECUTABLE=${prefix}/bin/git
+}
 
-    if {${os.major} < 9} {
-        #use fallback on Tiger since current version doesn't build and doxygen is usually just a build time dependency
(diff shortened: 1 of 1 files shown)


textproc/doxygen/Portfile
```diff
@@ -105,6 +105,20 @@ if {${os.platform} eq "darwin" && ${os.major} < 13} {
                         path:bin/git:git
     configure.args-append \
                         -DGIT_EXECUTABLE=${prefix}/bin/git
+
+    if {${os.major} < 9} {
```
barracuda156: Fine w/ me, but rather take this block out of `if {${os.platform} eq "darwin" && ${os.major} < 13} {`, make it separate.

barracuda156 (changes requested):

textproc/doxygen/Portfile
ForestExpertise: It is separate now, lmk if it is okay.

barracuda156 (approved):

PR #238, opened 2026-08-29
aqua/aqualink: bump to v0.5

watermark-hd:
Bumps aqua/aqualink from 0.4 to 0.5.

Fixes a connection failure specific to this port: devel/libsmb2 builds
with Kerberos/GSSAPI support by default, and libsmb2 itself defaults
to trying Kerberos before falling back to NTLM. Against a home NAS
with no krb5 realm/KDC, `gss_acquire_cred` fails and SPNEGO never
falls back to NTLM, so the connection is refused outright:

```
Connection failed: gss_acquire_cred: (Ein ungültiger Name wurde
übergeben., SPNEGO kann keine Mechanismen zum Aushandeln finden.)
```

AquaLink now explicitly requests NTLM authentication
(`smb2_set_authentication(ctx, SMB2_SEC_NTLMSSP)`), since it targets
home NAS / workgroup SMB shares rather than AD domains. Confirmed
building on real Tiger hardware; full writeup in `smb3/README.md`.

### DIFF
--- a/aqua/aqualink/Portfile
+++ b/aqua/aqualink/Portfile
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 PortGroup           makefile 1.0
 
 name                aqualink
-github.setup        watermark-hd ppc-mac-modernization 0.4 v
+github.setup        watermark-hd ppc-mac-modernization 0.5 v
 revision            0
 categories          aqua net
 license             unknown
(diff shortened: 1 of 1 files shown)


barracuda156 (approved):

barracuda156:
Thank you!

P. S. Plz consider setting some license. “Unknown” isn’t a valid value even )

PR #239, opened 2026-08-29
mesa: fix i386

alex-free:
This is the same as https://github.com/macos-powerpc/powerpc-ports/pull/221 but updated for new mesa version, as the original pull request got stale. Builds on Tiger i386, I do have to manually remove the problematic patch that @ForestExpertise 's pull request removes. However this pull request specifically doesn't remove that problematic patch in hopes that it will not conflict with https://github.com/macos-powerpc/powerpc-ports/pull/233 . If It does I can redo it.


alex-free:
Let me know if anything else needs to happen before this is merged @barracuda156

barracuda156 (approved):

PR #244, opened 2026-08-30
vvenc: fix i386 build, disable avx and sse4.2

alex-free:
I left this on for anything newer then snow leopard, and left it alone if not darwin since it was originally under a platform i386 block. 

The second cxxflag addition seems to be due to gcc16 stricter defaults with warnings so I added that where the other one was.

### DIFF
--- a/multimedia/vvenc/Portfile
+++ b/multimedia/vvenc/Portfile
@@ -33,14 +33,19 @@ configure.args-append \
                     -DVVENC_USE_ADDRESS_SANITIZER=OFF
 
 platform i386 {
-    configure.args-delete \
-                    -DVVENC_ENABLE_X86_SIMD=OFF
+    # This enables SSE4.2 (not availble on oldest intel CPUs) and AVX.
+    if {${os.platform} ne "darwin" || (${os.platform} eq "darwin" && ${os.major} > 10)} {        
+        configure.args-delete \
+            -DVVENC_ENABLE_X86_SIMD=OFF
+    }
 }
 
 compiler.cxx_standard   2014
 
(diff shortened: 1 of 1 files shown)


multimedia/vvenc/Portfile
```diff
@@ -33,14 +33,19 @@ configure.args-append \
                     -DVVENC_USE_ADDRESS_SANITIZER=OFF
 
 platform i386 {
-    configure.args-delete \
-                    -DVVENC_ENABLE_X86_SIMD=OFF
+    # This enables SSE4.2 (not availble on oldest intel CPUs) and AVX.
```
barracuda156, line 36: Note, `platform i386` is all x86, not only i386. Do we want not to have SIMD on 10.6 x86_64? (Not something I particularly care, but at least make sure this is actually what is intended.)

alex-free: 10.6 x86_64 *might* be able to do SSE4.2 if it's a 2009 or newer mac (Nehalem i7 came out towards then end of  2008 is when its first possible). But the issue I kinda saw with this is the main goal is to make binaries that actually run on all the early intel chips. So for example if a 2009 mac built this on 10.6 for x86_64, I couldn't run it on my 2007 C2D (which is 64 bit) on x86_64 snow leopard if SSE4.2 is on. AVX is 2011 and newer. And it's all or nothing so _just_ enabling SSE4.2 would require patches. These were really the main optimizations this turns on, it might make sense to go back later to see if any others are applicable but this makes sense to me for now. Perhaps some variant in the future +sse42 or something.

barracuda156: Ok, fine w/ me, I don’t care much, and the loss, if any, is minimal.

PR #252, opened 2026-08-31
tuntaposx: restore tiger compatibility

alex-free:
Uses last version that builds on tiger as per developer changelog: https://tuntaposx.sourceforge.net/download.xhtml.
It seems to be possible to build newer versions from say Leopard or Snow Leopard that target tiger according to that page (at least one version newer), but it doesn't help us _on_ tiger.

Trying to build the 'leopard' version (20111101) or any newer version then the one in the portfile requires apple GCC 4.2, and then is expecting leopard things like 'ld_classic' (which is present on newer Xcode then Tiger can run) when you pass it flags like -mkernel. This makes sense because all newer versions were built using newer Xcode/GCC.

 I'm pretty sure the macports built apple gcc 42 can't build kernel extensions for 10.4, at least not as-is without patching that compiler itself (the ld_classic thing is a leftover from macports 'porting' this from leopard to tiger, although that may not be the only issue as it still may not even work once that's resolved).

So this uses Xcode 2.5 gcc 4.0 (makes sense for kernel extensions). And luckily, the only patches needed were some reinplaces to change it from building universal -arch ppc -arch i386 to whatever is {configure.build_arch} (to match all other OS versions this supports).

Tested with openvpn2 and works great.

### DIFF
--- a/net/tuntaposx/Portfile
+++ b/net/tuntaposx/Portfile
@@ -17,9 +17,22 @@ long_description    Unix-style tun and tap virtual network interfaces for Mac \
 
 homepage            http://tuntaposx.sourceforge.net
 
+if {${os.major} == 8} {
+# Last version to build on tiger as per developer site.
+    version         20080804
+    revision        1
+    checksums       rmd160  b4fa7d3241d4b1ea97f3ca1db73f5669bc3d33ea \
+                    sha256  2cf027fb976dab6221071e9b8a85ef40cb867b065405e35bc823fdd9dba6fcc0 \
+                    size    33796
+
+    patchfiles      src_tap_makefile_tiger.diff \
+                    src_tun_makefile_tiger.diff
+
+    # error: jump to case label../tuntap.cc
+    configure.compiler gcc-4.0
+} elseif {${os.major} == 9} {
 # TODO: verify that a newer version works correctly on 10.6,
 # otherwise change the fallback threshold here.
-if {${os.major} < 10} {
     version         20111101
     revision        1
     checksums       rmd160  d7ec7fbea29d05de3a1532737b74970d9bbcc2ed \
@@ -27,7 +40,8 @@ if {${os.major} < 10} {
                     size    80453
 
     patchfiles      patch-src__tap__Makefile-leopard.diff \
-                    patch-src__tun__Makefile-leopard.diff
+                    patch-src__tun__Makefile-leopard.diff \
+                    patch-leopard.diff
 } else {
     version         20150118
     revision        1
@@ -65,10 +79,6 @@ if {${os.major} >= 18} {
             reinplace "s|@SDKROOT@|${configure.sdkroot}|" ${f}
         }
     }
-}
-
-if {${os.major} <= 9} {
-    patchfiles-append   patch-leopard.diff
 }
 
 use_configure       no
--- a/net/tuntaposx/files/src_tap_makefile_tiger.diff
+++ b/net/tuntaposx/files/src_tap_makefile_tiger.diff
@@ -1,0 +1,23 @@
+--- src/tap/Makefile.orig	2026-09-01 13:09:38.000000000 -0600
++++ src/tap/Makefile	2026-09-01 13:10:29.000000000 -0600
+@@ -23,16 +23,16 @@
+ 	-I/System/Library/Frameworks/Kernel.framework/Headers \
+ 	-I/System/Library/Frameworks/Kernel.framework/Headers/bsd
+ CFLAGS = -static -nostdinc -Wall -msoft-float -mlong-branch -force_cpusubtype_ALL -fno-builtin \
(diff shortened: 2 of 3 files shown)


net/tuntaposx/Portfile
```diff
@@ -44,12 +51,29 @@ distname            tuntap_${version}_src
...
 
-post-patch {
-    foreach {d} {tun tap} {
-        set f ${worksrcpath}/src/${d}/Makefile
-        reinplace "s|@CC@|${configure.cc}|g" ${f}
-        reinplace "s|@CXX@|${configure.cxx}|g" ${f}
-        reinplace "s|@ARCHFLAGS@|[get_canonical_archflags cc]|" ${f}
+# For 20080804 it does not use @CC@/@CXX@/@ARCHFLAGS@, so no need to match it.
+# In fact, since we use Xcode compilers, we can leave it.
+if {${os.major} < 8} {
```
barracuda156: This is wrong. Also, it should be in “if else” form and not two separate blocks.

barracuda156 (changes requested):

alex-free:
Implemented suggested changes, also used configure.compiler gcc-4.0 (much cleaner now)

net/tuntaposx/Portfile
```diff
@@ -44,12 +53,20 @@ distname            tuntap_${version}_src
...
-        reinplace "s|@ARCHFLAGS@|[get_canonical_archflags cc]|" ${f}
+        
+        if {${os.major} > 8} {            
+            reinplace "s|@CC@|${configure.cc}|g" ${f}
+            reinplace "s|@CXX@|${configure.cxx}|g" ${f}
+            reinplace "s|@ARCHFLAGS@|[get_canonical_archflags cc]|" ${f}
+        } elseif {${os.major} == 8} {
+            reinplace "s|-arch ppc -arch i386|-arch ${configure.build_arch}|g" ${f}
+            reinplace "s|gcc|${configure.cc}|g" ${f}
+            reinplace "s|g++|${configure.cxx}|g" ${f}
```
barracuda156: Reinplacing something as generic as `gcc` is unsafe. What we need is patching correct spots to `@CC@` etc. Which makes this change redundant (it is the _same_ reinplace), the only thing needed for 10.4 is a different patch.

Patches should all go into the block above, where checksums are. (No idea why a single patch is added separately below, but I can fix that in a follow-up, or you may fix it here itself.)

barracuda156 (changes requested):

net/tuntaposx/Portfile
alex-free: There is no ```@CC@```  in the 2008 version, which is why this is different. The sed could be i.e. CC = gcc or something to make it more verbose I guess but I don't really see why that would be better. Same with ```@CXX@``` and ```@ARCHFLAGS@``` that all got introduced in newer versions. Leaving it as g++/gcc in the makefile is problematic if you have done i.e. sudo port select --set gcc mp-gcc16

alex-free: Alternatively I could make patches to the makefiles to call /usr/bin/gcc-4.0 etc.

barracuda156: `@CC@` and friends are patched in, see https://github.com/macos-powerpc/powerpc-ports/blob/c6fe27026fdf295c2644242d626feeb5949da032/net/tuntaposx/files/patch-src-tun-Makefile.diff etc. Exactly the same is needed for 10.4.

P. S. Reinplacing `gcc` directly will replace all instances of this string, which may incidentally work, but is quite obviously wrong in principle.

alex-free: Ok I get it I'll make tiger patches then reinplace will be universal among all versions

barracuda156: P. S. And reinplacing `CC = gcc` or alike could be okay (we can be reasonably sure random lines are not changed), but here it is better to have identical logic across versions.

alex-free:
Fixed @barracuda156 . Also cleaned up stray leopard patch

net/tuntaposx/Portfile
```diff
@@ -17,17 +17,30 @@ long_description    Unix-style tun and tap virtual network interfaces for Mac \
...
+    version         20080804
+    revision        1
+    checksums       rmd160  b4fa7d3241d4b1ea97f3ca1db73f5669bc3d33ea \
+                    sha256  2cf027fb976dab6221071e9b8a85ef40cb867b065405e35bc823fdd9dba6fcc0 \
+                    size    33796
+
+    patchfiles      src_tap_makefile_tiger.diff \
+                    src_tun_makefile_tiger.diff
+
+    configure.compiler gcc-4.0
```
barracuda156, line 32: Looks good, could you just add a single-line comment why we need to hardcode compiler here? So that whoever reads portfile later does not need to dig into the history.

barracuda156 (approved):

barracuda156 (approved):

PR #253, opened 2026-09-01
ghostscript: fix i386 build. fix tiger build with gcc16.

alex-free:
I believe the gcc16 issue would impact all os versions as this occurs when using macports libiconv headers rather then system libiconv.

The i386 issue is old GCC4 versions don't have immintrin.h, this is an intel specific thing for SIMD like SSE2 (which we can actually use here as core solo/duo support up to SSE3).

### DIFF
--- a/print/ghostscript/Portfile
+++ b/print/ghostscript/Portfile
@@ -1,0 +1,240 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           muniversal 1.0
+
+name                ghostscript
+version             10.07.1
+revision            0
+
+categories          print
+license             AGPL-3 BSD
+maintainers         nomaintainer
+description         GPL Ghostscript, An interpreter for PostScript and PDF
+long_description    Ghostscript is the well-known PostScript interpreter which \
+                    is available for all common and most esoteric platforms and \
+                    supports many different printers and some displays.
+homepage            https://www.ghostscript.com/
+
+master_sites        https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs[strsed ${version} {g/\.//}]:source \
+                    sourceforge:project/gs-fonts/gs-fonts/6.0%20%28misc%2C%20AFPL%29/:otherfonts \
+                    sourceforge:project/gs-fonts/gs-fonts/8.11%20%28base%2035%2C%20GPL%29/:stdfonts \
+                    https://github.com/adobe-type-tools/mapping-resources-pdf/archive/:misc
+
+# Note: this needs to be manually updated for new upstream commits
+set mappingresources_commit \
+                    2dd5e53fb74a01718b9dfd448a0d1cce6fff2aa5
+
+distname            ghostpdl-${version}
+distfiles           ${distname}.tar.gz:source \
+                    ghostscript-fonts-other-6.0.tar.gz:otherfonts \
+                    ghostscript-fonts-std-8.11.tar.gz:stdfonts \
+                    ${mappingresources_commit}.zip:misc
+
+checksums           ${distname}.tar.gz \
+                    rmd160  f8d8fc6dfff453c97f565a33a335b05db39b474e \
+                    sha256  5c580ed888ce42ce4d76b8afac302e8b507e08d05e52e05b3649e0732559bfe4 \
+                    size    99962592 \
+                    ghostscript-fonts-other-6.0.tar.gz \
+                    rmd160  ab60dbf71e7d91283a106c3df381cadfe173082f \
+                    sha256  4fa051e341167008d37fe34c19d241060cd17b13909932cd7ca7fe759243c2de \
+                    size    796086 \
+                    ghostscript-fonts-std-8.11.tar.gz \
+                    rmd160  10a19a10d0388bc084a7c1d3da845068d7169054 \
+                    sha256  0eb6f356119f2e49b2563210852e17f57f9dcc5755f350a69a46a0d641a0c401 \
+                    size    3752871 \
+                    ${mappingresources_commit}.zip \
+                    rmd160  a65458ab9955421cf3085cf84a6eac299e8c93cb \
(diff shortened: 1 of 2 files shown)


print/ghostscript/Portfile
```diff
@@ -105,6 +105,18 @@ depends_run-append \
...
+# matching_tag_mask.h:15:23: error: immintrin.h: No such file or directory
+if {${configure.build_arch} eq "i386"} {
+    compiler.blacklist *gcc-4.*
+}
+
+# When linking macports libiconv:
+# gdevopvp.c:1685:39: error: passing argument 2 of 'libiconv' from incompatible
+# pointer type [-Wincompatible-pointer-types] 
+if {[string match "macports-gcc-*" ${configure.compiler}]} {
+    configure.cflags-append "-Wno-error=incompatible-pointer-types"
```
barracuda156: Why does this need to be quoted?

print/ghostscript/Portfile
```diff
@@ -105,6 +105,18 @@ depends_run-append \
 compiler.c_standard 1999
 configure.cflags-append -std=gnu99
 
+# matching_tag_mask.h:15:23: error: immintrin.h: No such file or directory
+if {${configure.build_arch} eq "i386"} {
```
barracuda156: Does this work for x86_64 though?

barracuda156:
@alex-free Could you parse ports for 10.5-10.6 ppc in your current infrastructure? This stream of imports from MacPorts solely for the sake of Tiger increases indexing burden for everyone else (in addition to increased maintenance burden for me).

print/ghostscript/Portfile
alex-free: It would have the same issue on i.e. x86_64 snow leopard. Would platform i386 be best?

print/ghostscript/Portfile
alex-free: will change, not sure why my file of common macports conditionals had this quoted.

print/ghostscript/Portfile
barracuda156: Yeah, I think so.

barracuda156: Hold on, that would unnecessarily include Rosetta.

Let’s keep `configure.build_arch` but add x86_64.

alex-free:
> @alex-free Could you parse ports for 10.5-10.6 ppc in your current infrastructure? This stream of imports from MacPorts solely for the sake of Tiger increases indexing burden for everyone else (in addition to increased maintenance burden for me).

Like provide portindexes? That's just a matter of editing https://github.com/alex-free/tigerports-infrastructure/blob/0439e0c5f385ac94529a46f606d5cf43090606fc/jobs/mprsyncup#L78 . This is how I get around needing to portindex on local hardware (I offload it to the server like macports)

print/ghostscript/Portfile
alex-free: sure

alex-free:
Also I'll try to help maintain the fixes I submit, I'm going to create a list to keep track of them on tigerports.com (and upstream status when applicable).

alex-free:
hold up on this, I'm going to change it

alex-free:
Ok good to go.

alex-free:
@barracuda156 can we merge this here too?

print/ghostscript/Portfile
```diff
@@ -105,6 +105,19 @@ depends_run-append \
 compiler.c_standard 1999
 configure.cflags-append -std=gnu99
 
+# Older gcc 4 does not provide the header.
+# matching_tag_mask.h:15:23: error: immintrin.h: No such file or directory
+if {${configure.build_arch} eq "i386" || ${configure.build_arch} eq "x86_64"} {
```
barracuda156, line 110:
```
if {${configure.build_arch} in [list i386 x86_64]} {
```

P. S. This (usage of `${configure.build_arch}`) probably won’t work with universal builds, but given that universal builds involving ppc are largely broken anyway, I don’t really care.

print/ghostscript/Portfile
```diff
@@ -105,6 +105,19 @@ depends_run-append \
...
+    compiler.blacklist *gcc-4.*
+}
+
+# When including macports libiconv:
+# gdevopvp.c:1685:39: error: passing argument 2 of 'libiconv' from incompatible
+# pointer type [-Wincompatible-pointer-types] 
+if {[string match macports-gcc-* ${configure.compiler}]} {
```
barracuda156: This needs verification on 10.5/10.6.

barracuda156 (changes requested):
I will check on 10.6 when I have time (will try doing it soon), but please change arch condition here.

print/ghostscript/Portfile
alex-free: Good idea, it could be how macports libiconv itself is interacting with system headers. I'm going to set up newer OS installs I just don't want to be stretched too thin.

barracuda156: @alex-free I confirmed successful build on 10.6 with gcc-16 without adding this flag (or any other changes in portfile). So this should not be added unconditionally, it is a Tiger-specific issue.

alex-free:
If there’s better arch detection that preserves Rosetta, let me know

print/ghostscript/Portfile
```diff
@@ -0,0 +1,255 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           muniversal 1.0
+
+name                ghostscript
+version             10.07.1
```
barracuda156, line 7: This needs to use the current version: https://github.com/macports/macports-ports/blob/3e41875ed2aa31f40f59edbb593ee11dffa61f51/print/ghostscript/Portfile

barracuda156 (changes requested):

barracuda156:
@alex-free I have imported the port in its current ver: https://github.com/macos-powerpc/powerpc-ports/commit/9b94740b8e6d3958d4c0d7b4c45b2213e0c6e71f

alex-free:
> @alex-free I have imported the port in its current ver: https://github.com/macos-powerpc/powerpc-ports/commit/9b94740b8e6d3958d4c0d7b4c45b2213e0c6e71f

I'll redo the patches thanks

alex-free:
redid new pull request for imported ver

PR #254, opened 2026-09-01
gjs: fix Tiger build

ForestExpertise:
May be worth checking if fpermissive is needed on Leopard as well.

### DIFF
--- a/gnome/gjs/Portfile
+++ b/gnome/gjs/Portfile
@@ -104,3 +104,10 @@ if {${os.platform} eq "darwin" && ${os.major} < 16} {
 # at the system level during testing.
 test.run            yes
 test.target         test
+
+platford darwin 8 {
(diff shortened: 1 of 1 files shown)


gnome/gjs/Portfile
```diff
@@ -104,3 +104,10 @@ if {${os.platform} eq "darwin" && ${os.major} < 16} {
 # at the system level during testing.
 test.run            yes
 test.target         test
+
+platford darwin 8 {
+    configure.cxxflags-append \
+                            -fpermissive
+    configure.cflags-append \
+                            -Wno-error=incompatible-pointer-types
```
barracuda156, line 112: @ForestExpertise Nitpicking, but could you align with text above? (Or at least do not make a redundant line break for cflags.)

barracuda156 (changes requested):

PR #255, opened 2026-09-01
aqua/aqualink: bump to v0.5.3

watermark-hd:
Bumps aqua/aqualink from 0.5.2 to 0.5.3.

While diagnosing a real auth error reported against a build using this
port's libsmb2, found that connection-failure messages weren't showing
their detail line at all — the status field displaying them is a
single-line NSTextField, so the second line (the actual underlying
error text) was simply invisible, not truncated. This blocked getting
the information needed to diagnose the report properly.

v0.5.3 also shows the full message in a dialog. Confirmed on real
Tiger hardware. See smb3/README.md.

### DIFF
--- a/aqua/aqualink/Portfile
+++ b/aqua/aqualink/Portfile
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 PortGroup           makefile 1.0
 
 name                aqualink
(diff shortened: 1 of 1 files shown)


barracuda156 (approved):
I assume it was tested, gonna merge

PR #264, opened 2026-09-04
orc: tiger ppc fix, comment patches for intel

alex-free:


devel/orc/Portfile
```diff
@@ -38,13 +38,29 @@ platform darwin 8 {
...
 }
 
 if {${configure.build_arch} eq "i386"} {
+    # It is not enough to disable the optimizations as it still tries to use
+    # instructions too new to identify them when detecting anything intel.
+    # orccpu-x86.c {standard input}:15:no such instruction: `xgetbv'
     patchfiles-append \
         orc_orccpu-x86.c_i386.diff
+} elseif {${configure.build_arch} eq "ppc"} {
+    # Disables SSE/AVX/MMX/everything when on PPC.
```
barracuda156: It should not be needed, since I build it on ppc without passing this argument. (I can easily imagine it _could_ be needed, just de facto it works without it.)
If this actually desirable, then perhaps ppc64 should join in. But I need to see what this option actually entails before considering merging.

devel/orc/Portfile
```diff
@@ -38,13 +38,29 @@ platform darwin 8 {
...
+    configure.args-append \
+        -Dorc-target=
+}
+
+variant g4 {
+    configure.args-replace \
+        -Dorc-target= \
+        -Dorc-target=altivec
```
barracuda156: Why should this be specific to G4? 

P. S. On a side note, we need consistent naming, so since `G4` is capitalized elsewhere in variants, that should be used. But I am not convinced cpu-specific variant is justified here at all.

barracuda156 (changes requested):

devel/orc/Portfile
alex-free: This should actually be in the darwin 8 block now that I think about it. The whole issue is by default it still builds x86 specific files, and I used a intel-specific built-in to overcome tiger not having a specific flush operation in system headers. So the idea was to avoid having it build the x86 files at all, because theres not an equievelnt builtin for ppc (and it's x86 files anyways). So on 10.6, it does have the system headers capable of the flush operation it needs.

alex-free: Can still do ppc64 (as eventually it would be cool to see how broken that is with tiger server )

devel/orc/Portfile
alex-free: Was being pedantic, this has the same side effect as -Dorc-target but noticing the altivec target also accomplished not building the x86 specific files I do not have an equievelnt gcc builtin for ppc for (see above) I didn't want to accidentally break G3 builds if this would. I don't have a G3 to find out. If this could possibly break G3. Due to this, it should actually also be in the darwin8 block too. And I can change the name capitalization too I didn't realize. If there is a better name then G4 I can do that too.

barracuda156: There is a configure-time check for altivec support, see ffmpeg ports, for example. I am not aware of any actual G3 users of our ports, and G3 is an unsupported cpu for an unsupported OS (I guess no 10.5 on G3?), so not really a concern, at least not until someone complains, but altivec works on G5, so it looks odd to single out G4 here. It is also undesirable to have per-cpu variants without a very good reason to have them (and unless optimizations are for-cpu, there is no reason to have any, IMO just enable altivec, if is it disabled by default).

alex-free: Thats fine will remove.

barracuda156: To be clear, it is _better_ not to break something for G3 if that takes not much extra effort, just not a requirement for a PR (until someone shows up with a G3 and a willingness to test stuff on it). On the contrary, anything done for ppc/ppc64 that affects 10.5/10.6 should be done carefully – we do not want to break anything for primary systems.

alex-free: How this build system works is everything is always on (so altivec is 'on' with intel too, AVX is 'on' with ppc though obviously not actually). However specifying altivec explicitly prevents every other thing (mmx/sse/avx) from being on when NOT built on x86 or x86_64 so it doesn't even parse the x86 specific files.

I'm going to write a note on how to make it work on g3 if it even is an issue there, we don't have a way to test anyways.

barracuda156: I rather keep the default then, since altivec is enabled anyway, but passing it gonna produce adverse effects on universal builds for no good reason. Feel free to make needed adjustment inside `darwin 8` block.

alex-free:
Anything that can be done to improve this?

alex-free:
This now checks for altivec @barracuda156 so it should be safe for G3. Uses same check as ffmpeg, so all ppc and intel now work once this is merged

PR #269, opened 2026-09-06
libfmt12: undef tiger sdk macro before base.h macro clashes

alex-free:
In ctypes.h of tiger sdk, there exists a macro of the same name with different signature that is unrelated, which for example when using this header while building doxygen on tiger causes the build to fail

### DIFF
--- a/devel/libfmt12/files/include_fmt_base.h_tiger.diff
+++ b/devel/libfmt12/files/include_fmt_base.h_tiger.diff
@@ -1,0 +1,20 @@
+--- include/fmt/base.h.orig	2026-09-10 20:05:17.000000000 -0600
++++ include/fmt/base.h	2026-09-10 20:05:33.000000000 -0600
+@@ -891,6 +891,17 @@
+ #  define FMT_USE_LOCALE (FMT_OPTIMIZE_SIZE <= 1)
+ #endif
+ 
++// Mac OS X 10.4u SDK has conflicting isalpha macro from ctype.h that has a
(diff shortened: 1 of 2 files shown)


devel/libfmt12/files/include_fmt_base.h_tiger.diff
```diff
@@ -0,0 +1,14 @@
+--- include/fmt/base.h.orig	2026-09-05 21:44:55.000000000 -0600
++++ include/fmt/base.h	2026-09-05 21:47:44.000000000 -0600
+@@ -891,6 +891,11 @@
+ #  define FMT_USE_LOCALE (FMT_OPTIMIZE_SIZE <= 1)
+ #endif
+ 
++// Mac OS X 10.4 has name clash: isalpha(c) __istype((c), _CTYPE_A)
++#ifdef isalpha
```
barracuda156: Please make this conditional here as well (it is safer, and doing it is a trivial copypaste of known macro plus header).

barracuda156 (changes requested):

devel/libfmt12/files/include_fmt_base.h_tiger.diff
barracuda156: @alex-free This is the only stopper

alex-free: No worries it’s better with the macro I agree. It’s the next thing I’m getting to

alex-free:
Redid patch for only below 10.5, matched indentation/code style

barracuda156 (approved):

PR #270, opened 2026-09-06
doxygen: fix tiger fallback checksums

alex-free:
The issue here is the distfile/version is set well before we get to where it is changed currently for tiger, this is the cleanest way to fix it and mirrors tuntaposx pr.

This in combination with the libfmt12 pr restores doxygen for tiger

### DIFF
--- a/textproc/doxygen/Portfile
+++ b/textproc/doxygen/Portfile
@@ -21,6 +21,20 @@ version                 1.18.0
 revision                0
 epoch                   2
 
+checksums               rmd160  91f625d6fcfe87a511ef1e356e7fa14f5dcaf190 \
+                        sha256  a1deed70a6785bbec95a2b2a9e419dc7f7b223a9d74a8644ae611c8e2dcdd354 \
+                        size    9905783
+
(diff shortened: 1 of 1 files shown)


textproc/doxygen/Portfile
```diff
@@ -21,6 +21,26 @@ version                 1.18.0
 revision                0
 epoch                   2
 
+if {${os.platform} eq "darwin" && ${os.major} < 9} {
```
barracuda156, line 28: Not in this sequence at least. Place modern systems first, please.

barracuda156 (changes requested):

barracuda156 (approved):
Looks fine, just to confirm, did you verify that it works after `sudo port clean --all doxygen`? (So that it does not use earlier downloaded tarball, but verifies this end-to-end.)

alex-free:
> Looks fine, just to confirm, did you verify that it works after `sudo port clean --all doxygen`? (So that it does not use earlier downloaded tarball, but verifies this end-to-end.)

yup, fresh download + checksums correct.

barracuda156 (approved):

PR #271, opened 2026-09-06
minizip-ng: don't mandate O_NOFOLLOW for tiger and lower

alex-free:
While this is a platform darwin 8 block here, it technically fixes it for lower as well once I get it to upstream

### DIFF
--- a/archivers/minizip-ng/Portfile
+++ b/archivers/minizip-ng/Portfile
@@ -1,0 +1,82 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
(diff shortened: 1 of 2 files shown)


archivers/minizip-ng/Portfile
```diff
@@ -0,0 +1,90 @@
...
+    configure.args-replace \
+                    -DMZ_ZLIB=OFF       -DMZ_ZLIB=ON
+}
+
+# Adds support for not having O_NOFOLLOW.
+platform darwin 8 {
+    patchfiles-append \
+        mz_zip_rw.c_tiger.diff \
+        mz_strm_os_posix.c_tiger.diff \
+        mz.h_tiger.diff
```
barracuda156: Could we have this in a single patch?

barracuda156 (changes requested):

alex-free:
All one patch

barracuda156 (approved):

PR #272, opened 2026-09-08
expat: remove rpath from cmake config on tiger

alex-free:
When opencolorio tries to use expat on tiger, configuration fails due to the @rpath 

CMake Error at /opt/local/lib/cmake/expat-2.8.4/expat.cmake:59 (add_library):
  Attempting to use @rpath without CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG being
  set. This could be because you are using a Mac OS X version less than 10.5
  or because CMake's platform configuration is corrupt.

So on 10.4, reinplace the rpath with an absolute path based on prefix

### DIFF
--- a/textproc/expat/Portfile
+++ b/textproc/expat/Portfile
@@ -23,6 +23,11 @@ checksums           rmd160  e90e6523b6b69a3969502b1276635e16a376c48d \
                     sha256  963250a823c16a498582b4ad82ad0f88926be0769675d3b6956be4d769a1cd8f \
                     size    666106
 
+# Signify rpath fix.
+platform darwin 8 {
+    incr revision
+}
+
 # Don't use a compression format that necessitates a dependency on another
 # port since that causes a dependency cycle on older OS versions since
 # e.g. clang-3.4 depends on python27-bootstrap which depends on expat.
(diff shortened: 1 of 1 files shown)


textproc/expat/Portfile
```diff
@@ -23,6 +23,11 @@ checksums           rmd160  e90e6523b6b69a3969502b1276635e16a376c48d \
                     sha256  963250a823c16a498582b4ad82ad0f88926be0769675d3b6956be4d769a1cd8f \
                     size    666106
 
+# Signify rpath fix.
+platform darwin 8 {
+    revision 1
```
barracuda156: @alex-free Thank you. I am fine with the change, but please use `incr revision` here (that way if port is revbumped, Tiger gets +1 automatically) and preferably also change the note (or add to) to “drop on next update”, so that it is visible.

textproc/expat/Portfile
```diff
@@ -62,6 +67,11 @@ if {${os.platform} eq "darwin" && ${os.major} < 13} {
 }
 
 post-destroot {
+    # If expat has an @rpath, it will fail on 10.4 when checked by cmake.
+    # Simply remove the @rpath and replace it with an absolute path.
+    if {${os.platform} eq "darwin" && ${os.major} == 8} {
```
barracuda156: Preferred form is either `platform darwin 8` (easier to read) or `${os.platform} eq "darwin" && ${os.major} < 9` (assuming the fix is not really exclusive to 10.4). With `${os.major} == 8` it is both less readable and less logically correct.

barracuda156 (changes requested):

textproc/expat/Portfile
alex-free: I couldn’t for the life of me remember what the syntax was for that. Will change

barracuda156 (approved):

PR #273, opened 2026-09-08
mplayer: use correct sdl dep

alex-free:
mplayer uses sdl1, not sdl2. So when building this, it pulls in libsdl2 but then never finds sdl.

Curiously, official macports disables sdl: https://github.com/macports/macports-ports/blob/db439281734923e731f8c6a86bb6631d74bdbd45/multimedia/MPlayer/Portfile#L66

Apparently, libsdl port is now a wrapper for sdl2 for newer operating systems: https://github.com/macports/macports-ports/blob/master/devel/libsdl/Portfile but for older ones it uses real sdl1 (and tiger still works).

Sure enough, trying to build libsdl12-compat fails on tiger due to objc abi (the above stays working however because it uses actual sdl1 on older operating systems). 

By enabling sdl, OTB we get a 1080p playback  on G5 tiger (pretty competent anyways, some minor slowdown with youtube H264 rips but hey, powervlc does that too and I haven't even upgraded my GPU yet in my early 05 dual 2ghz nor tried the 'G5' variant, or done much other additional flags/config).

The best thing to do here is mplayer -vo gl_nosw -ao sdl on G5 tiger for performance (-vo sdl is much too slow as it's to my understanding software based do to using actual sdl1 port on tiger). SDL is what enables the various gl based vos btw, there is currently zero video outs available if you didn't happen to install libsdl.

Hope this makes sense! @ForestExpertise may be interested in this as well.

### DIFF
--- a/multimedia/MPlayer/Portfile
+++ b/multimedia/MPlayer/Portfile
@@ -47,7 +47,7 @@ depends_lib-append  port:aom \
                     port:libxml2 \
                     port:lzo2 \
                     port:ncurses \
-                    port:libsdl2 \
+                    port:libsdl \
                     path:lib/libspeex.dylib:speex \
                     port:zlib
 


ForestExpertise:
> mplayer uses sdl1, not sdl2. So when building this, it pulls in libsdl2 but then never finds sdl.
> 
> Curiously, official macports disables sdl: https://github.com/macports/macports-ports/blob/db439281734923e731f8c6a86bb6631d74bdbd45/multimedia/MPlayer/Portfile#L66
> 
> Apparently, libsdl port is now a wrapper for sdl2 for newer operating systems: https://github.com/macports/macports-ports/blob/master/devel/libsdl/Portfile but for older ones it uses real sdl1 (and tiger still works).
> 
> Sure enough, trying to build libsdl12-compat fails on tiger due to objc abi (the above stays working however because it uses actual sdl1 on older operating systems).
> 
> By enabling sdl, we get a 1080p playback on G5 tiger (pretty competent anyways, some minor slowdown with youtube H264 rips but hey, powervlc does that too and I haven't even upgraded my GPU yet in my early 05 dual 2ghz nor tried the 'G5' variant).
> 
> The best thing to do here is mplayer -vo gl_nosw -ao sdl on G5 tiger for performance (-vo sdl is much too slow as it's to my understanding software based do to using actual sdl12 port on tiger). SDL is what enables the various gl based vos btw, there is currently zero video outs available if you didn't happen to install libsdl.
> 
> Hope this makes sense! @ForestExpertise may be interested in this as well.

Great find! I had just assumed it used sdl2. I had both active, so maybe it was finding the correct sdl anyway. I am excited to see performance increases with sdl, or at least with the suggested vos.
You may consider adding a rev bump to force rebuilds, though since I would be surprised if someone other than me has mplayer installed. Then again, if the theory is that my systems are already using sdl, not sdl2, this is unnecessary. If @barracuda156 doesn't want a rev bump, I will probably uninstall and reinstall to make sure it built with the correct sdl.

barracuda156:
Looks like SDL2 is indeed wrong here, but it is not evident if libsdl or libsdl12-compat is preferable.
https://openmamba.org/en/rpms/base/mplayer
vs
https://codeberg.org/guix/guix/src/commit/f707b24d86af2375fc35fb3d6f9959f6f17c225c/gnu/packages/video.scm#L2880

alex-free:
> Looks like SDL2 is indeed wrong here, but it is not evident if libsdl or libsdl12-compat is preferable. https://openmamba.org/en/rpms/base/mplayer vs https://codeberg.org/guix/guix/src/commit/f707b24d86af2375fc35fb3d6f9959f6f17c225c/gnu/packages/video.scm#L2880

If you want to change for say 10.6 if it uses libsdl2-compat or actual libsdl, it makes sense to modify the libsdl port that decides that based on os ver: https://github.com/macports/macports-ports/blob/b414725d040d690cbcefd6964ce0bae2044a870f/devel/libsdl/Portfile#L38

That way it is a global change for anything libsdl (which admittedly is probably not much anymore)

Perhaps libsdl2-compat enables some kind of hardware acceleration translation or better performance, I'm not too sure (but it's possible)

barracuda156 (approved):
I will just merge this: libsdl2 is admittedly wrong in this case, so nothing to lose. Not interested to test this personally or deal with sdl1 in any way.

PR #277, opened 2026-09-10
libsdl2: more agressive/optimized buffer patch from cygwin for pulseaudio variant(s)

alex-free:
Pulseaudio on audio device open implemented by SDL2 waits for it's buffer to fill with sound data before it starts playing (to conservatively prevent buffer underruns). When using it directly as an audio out with SDL (tested with ffplay as that's the easist thing to build right now under tiger) this is noticeable as there is an audible delay until audio starts. This desyncs video playback with a ~1 second audio latency delay, lip syncing is off, etc. I noticed that when switching pulse audio sinks (internal speaker/headphone to my usb sound card to bluetooth adapter) back and forth it 're-negotiated' the latency/buffer sync and fixed the delay.

This [patch is from cygwin](https://cygwin.com/git/?p=git/cygwin-packages/SDL2.git;a=commitdiff;h=83a1059d34f39abd3f0e8cfd5bb8507152f00d80;ds=inline), where they have SDL also using pulseaudio as an audio out. And after much testing it essentially does the behavior that switching sinks does by default when starting an audio stream. I did test this for over a day with a ton of different 720p YouTube videos with FFplay just to make sure this info above is accurate. I also did not encounter any buffer underuns during any testing, coreaudio feeds pulse data fast enough consistently.

When detecting pulseaudio variant I have done incr revision to signify SDL2 stuff should be rebuilt.

Since more SDL2 stuff builds on 10.6 currently, any more testing/comparison pre-patch/post-patch would be appreciated

### DIFF
--- a/devel/libsdl2/Portfile
+++ b/devel/libsdl2/Portfile
@@ -193,6 +193,11 @@ variant pulseaudio description "Build with PulseAudio support" {
 
     configure.args-replace \
                     --disable-pulseaudio --enable-pulseaudio
+
+    # Smaller more agressive buffers like Cygwin to fix initial audio delay on 
+    # audio device open to fill buffer and sync issues. Tested with FFplay.
+    incr revision
+    patchfiles-append    src_pulseaudio_sdl_pulseaudio.c_trim_buffers.diff
 }
 
 variant samplerate description "Build with libsamplerate support" {
--- a/devel/libsdl2/files/src_pulseaudio_sdl_pulseaudio.c_trim_buffers.diff
+++ b/devel/libsdl2/files/src_pulseaudio_sdl_pulseaudio.c_trim_buffers.diff
@@ -1,0 +1,13 @@
+--- src/audio/pulseaudio/SDL_pulseaudio.c.orig	2026-09-09 10:13:30.000000000 -0600
++++ src/audio/pulseaudio/SDL_pulseaudio.c	2026-09-09 10:24:40.000000000 -0600
+@@ -663,8 +663,8 @@
+     /* Reduced prebuffering compared to the defaults. */
+     paattr.fragsize = this->spec.size;
+     paattr.tlength = h->mixlen;
+-    paattr.prebuf = -1;
+-    paattr.maxlength = -1;
++    paattr.prebuf = iscapture ? -1 : 0;
++    paattr.maxlength = iscapture ? -1 : (h->mixlen * 2);
+     paattr.minreq = -1;
+     flags |= PA_STREAM_ADJUST_LATENCY;
+ 


barracuda156:
@alex-free Any idea why they deleted that patch though? Newer version does not have that part anymore: https://cygwin.com/git/?p=git/cygwin-packages/SDL2.git;a=blob;f=SDL2-2.30.3-2.src.patch;h=44a7e818a56e0bc02b625a24f6004ec2d3542200;hb=83a1059d34f39abd3f0e8cfd5bb8507152f00d80
Also, in any case we want the link to the source of the patch (perhaps in patch file rather than portfile).

alex-free:
> @alex-free Any idea why they deleted that patch though? Newer version does not have that part anymore: https://cygwin.com/git/?p=git/cygwin-packages/SDL2.git;a=blob;f=SDL2-2.30.3-2.src.patch;h=44a7e818a56e0bc02b625a24f6004ec2d3542200;hb=83a1059d34f39abd3f0e8cfd5bb8507152f00d80 Also, in any case we want the link to the source of the patch (perhaps in patch file rather than portfile).

Yea I can do that. Honestly not sure but from a glimpse (there is a severe lack of comments/commit messages) it seems like they are handling audio differently then they used to before moving entirely to SDL3.

I can reach out to the maintainer there, that's probably the only way to get more info surrounding this.

alex-free:
@barracuda156 I've been looking into this more. 

Relevant is the pulseaudio code is the same that the cygwin patch modifies in SDL3 devel:
https://github.com/libsdl-org/SDL/blob/22245316bf41514b939270ec341e3f6ffc35fa18/src/audio/pulseaudio/SDL_pulseaudio.c#L746

So one thing I'm going to test is _only modifying the buffer length_:
paattr.maxlength = iscapture ? -1 : (h->mixlen * 2);

Notice this pulse audio doc (which is how SDL pulse audio is setup): https://www.freedesktop.org/software/pulseaudio/doxygen/structpa__buffer__attr.html#abef20d3a6cab53f716846125353e56a4

```
maxlength
uint32_t pa_buffer_attr::maxlength
Maximum length of the buffer in bytes.

Setting this to (uint32_t) -1 will initialize this to the maximum value supported by server, which is recommended.

In strict low-latency playback scenarios you might want to set this to a lower value, likely together with the PA_STREAM_ADJUST_LATENCY flag. If you do so, you ensure that the latency doesn't grow beyond what is acceptable for the use case, at the cost of getting more underruns if the latency is lower than what the server can reliably handle.
```

So actually, we shouldn't need to modify the prebuffer state (`paattr.prebuf = iscapture ? -1 : 0;` at all as that doesn't impact active latency, in fact I think the cywin patch was implemented wrong because the patch from cygwin does not do the PA_STREAM_START_CORKED for a paattr.prebuf being 0 according to the same pulseaudio docs: 

```
prebuf
uint32_t pa_buffer_attr::prebuf
playback only: pre-buffering.

The server does not start with playback before at least prebuf bytes are available in the buffer. It is recommended to set this to (uint32_t) -1, which will initialize this to the same value as tlength, whatever that may be.

Initialize to 0 to enable manual start/stop control of the stream. This means that playback will not stop on underrun and playback will not start automatically, instead [pa_stream_cork()](https://www.freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a14e698233ac2d246646651955ab0ec7b) needs to be called explicitly. If you set this value to 0 you should also set PA_STREAM_START_CORKED. Should underrun occur, the read index of the output buffer overtakes the write index, and hence the fill level of the buffer is negative.

Start of playback can be forced using [pa_stream_trigger()](https://www.freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#ae17a3a9f6ee0403c4665f6f4ce02ca3c) even though the prebuffer size hasn't been reached. If a buffer underrun occurs, this prebuffering will be again enabled.
```

The theory right now, is it's actually the size of the buffer being massive that receives coreaudio data. Pulseaudio keeps backing this up the more I look into it: https://freedesktop.org/software/pulseaudio/doxygen/streams.html

```
Buffer Attributes

Playback and record streams always have a server-side buffer as part of the data flow. The size of this buffer needs to be chosen in a compromise between low latency and sensitivity for buffer overflows/underruns.

The buffer metrics may be controlled by the application. They are described with a [pa_buffer_attr](https://freedesktop.org/software/pulseaudio/doxygen/structpa__buffer__attr.html) structure.

If PA_STREAM_ADJUST_LATENCY is set, then the tlength/fragsize parameters of the [pa_buffer_attr](https://freedesktop.org/software/pulseaudio/doxygen/structpa__buffer__attr.html) structure will be interpreted slightly differently than otherwise when passed to [pa_stream_connect_record()](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#abfd34293aae8f170f572d1093c1bcdf9) and [pa_stream_connect_playback()](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#ab9544f6677af133fbe81bf8a21eb489c): the overall latency that is comprised of both the server side playback buffer length, the hardware playback buffer length and additional latencies will be adjusted in a way that it matches tlength resp. fragsize. Set PA_STREAM_ADJUST_LATENCY if you want to control the overall playback latency for your stream. Unset it if you want to control only the latency induced by the server-side, rewritable playback buffer. The server will try to fulfill the client's latency requests as good as possible. However if the underlying hardware cannot change the hardware buffer length or only in a limited range, the actually resulting latency might be different from what the client requested. Thus, for synchronization clients always need to check the actual measured latency via [pa_stream_get_latency()](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#aa521efcc16fe2abf0f8461462432ac16) or a similar call, and not make any assumptions about the latency available. The function [pa_stream_get_buffer_attr()](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a9a3c3e78eafb28cce3a16cef2b68a385) will always return the actual size of the server-side per-stream buffer in tlength/fragsize, regardless whether PA_STREAM_ADJUST_LATENCY is set or not.

The server-side per-stream playback buffers are indexed by a write and a read index. The application writes to the write index and the sound device reads from the read index. The read index is increased monotonically, while the write index may be freely controlled by the application. Subtracting the read index from the write index will give you the current fill level of the buffer. The read/write indexes are 64bit values and measured in bytes, they will never wrap. The current read/write index may be queried using [pa_stream_get_timing_info()](https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a090147751441a97e04a4acef1d6514cb) (see below for more information). In case of a buffer underrun the read index is equal or larger than the write index. Unless the prebuf value is 0, PulseAudio will temporarily pause playback in such a case, and wait until the buffer is filled up to prebuf bytes again. If prebuf is 0, the read index may be larger than the write index, in which case silence is played. If the application writes data to indexes lower than the read index, the data is immediately lost.
```

If you look at the pulse audio port, there is actually a ton of patches to bridge core audio into it (and crazy enough, there is a tiger specific one (which well, makes sense because audio unit vs audio que etc.).

Pulse audio may be reporting a very large buffer/latency, more so then it needs. So it makes sense:

Fix pulse audio backend buffer
Agressively keep the one here

Still looking into it, but need to write this down somewhere

alex-free:
Ok, I am having good results modiying pulse audio buffer itself on intel to not include accounting for the fixed sink latency (core audio is not setup for variable latency like linux, it is dumb so it just sets it to 250ms, which makes the buffer massive, then applications typically do twice the size of this buffer). I'm going to rebuild on G5, test there, and report back. Modifying pulse audio is better because _anything_ that uses pulseaudio won't have this issue, it really just is expanding the patch set for coreaudio device. And best of all, we can keep the buffer sizes the same that actual applications use/don't have to potentially do fixes per app.

PR #281, opened 2026-09-11
cairo: fix distfile fetch

alex-free:
This doesn't work anymore because it's fetching html containing a redirect. If I remove this second line (like MacPorts has it right now) it does work: https://github.com/macports/macports-ports/blob/8949faaf0324208a6e920fa721a9576082c00fe1/graphics/cairo/Portfile#L29


barracuda156 (approved):
Odd that a fallback URL matters at all, but we can grant MacPorts that much, at least fetching is expected to work.

PR #282, opened 2026-09-12
xclock: use -f for autoreconf

alex-free:
I checked macports-ports and they [don't have any autoreconf](https://github.com/macports/macports-ports/blob/master/x11/xclock/Portfile) once downloaded, if I do that I get: 
```
SOURCE_DATE_EPOCH='1789255477'
Executing:  cd "/opt/local/var/macports/build/xclock-236e6181/work/xclock-1.1.1" && ./configure --prefix=/opt/local 
DEBUG: system:  cd "/opt/local/var/macports/build/xclock-236e6181/work/xclock-1.1.1" && ./configure --prefix=/opt/local 
sh: line 1: ./configure: No such file or directory
Command failed:  cd "/opt/local/var/macports/build/xclock-236e6181/work/xclock-1.1.1" && ./configure --prefix=/opt/local 
Exit code: 127
Error: Failed to configure xclock: configure failure: command execution failed
DEBUG: Error code: NONE
DEBUG: Backtrace: configure failure: command execution failed
    while executing
"$procedure $targetname"
Error: See /opt/local/var/macports/logs/_opt_local_var_macports_sources_tigerports.com_macports_release_x11_xclock/xclock/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets if you believe there
is a bug.
Error: Processing of port xclock failed
g5:~ alex$ 
```
So I assume this is something you added since your port is newer. Anyways specifically this is referring config.rpath so this is _probably_ tiger specific.

### DIFF
--- a/x11/xclock/Portfile
+++ b/x11/xclock/Portfile
@@ -22,6 +22,7 @@ checksums           rmd160  8bcfae78d8631389c5efeb8b32c988c6ee6f2fca \
                     size    31735
 
 use_autoreconf      yes
+autoreconf.args     -fiv
 
 depends_build-append \
                     path:bin/pkg-config:pkgconfig \


x11/xclock/Portfile
```diff
@@ -23,6 +23,13 @@ checksums           rmd160  8bcfae78d8631389c5efeb8b32c988c6ee6f2fca \
 
 use_autoreconf      yes
 
+# autopoint: File config.rpath has been locally modified.
+# autopoint: *** Some files have been locally modified. Not overwriting them
+# because --force has not been specified
```
barracuda156: This should not break anything.

alex-free: It stops the build because auto reconf does not return 0

x11/xclock/Portfile
```diff
@@ -23,6 +23,13 @@ checksums           rmd160  8bcfae78d8631389c5efeb8b32c988c6ee6f2fca \
...
+platform darwin 8 {
+    autoreconf.args     --force --install
```
barracuda156: If this is needed, use it unconditionally and drop the comment (it does not explain the error).
`-fiv` is a common set of args.

barracuda156 (changes requested):
If it works w/out autoreconf, it can be dropped, otherwise no need for a special case, it is a standard thing.

x11/xclock/Portfile
barracuda156: I mean that message is not the error (at least I have seen it during normal operation).

alex-free:
Are you sure? The issue is there is no configure in tarball, so it refuses to write the configure. this might've been non-fatal if there had been a configure already (in other ports) but I suspect it was because of it detecting lack of rpath on tiger

alex-free:
Dropped comment, does shorthand -fiv args unconditionally

barracuda156 (approved):

PR #284, opened 2026-09-13
cairo: remove hack, proper quartz support on tiger

alex-free:
cairo has a long history of getting back 10.4 support then dropping it: https://cairo.freedesktop.org/news/cairo-1.16.0/
so upstream should be supportive if they know there are users as they last supported it in 2018. I used old cairo versions to base this off of. Basically 10.4 API is different then 10.5+ as per usual.

This gets us back to proper variant quartz+x11 like macports upstream on all platforms including darwin 8. This also fixes py-cairo+x11+xquartz issue.

### DIFF
--- a/graphics/cairo/Portfile
+++ b/graphics/cairo/Portfile
@@ -115,12 +115,15 @@ platform macosx {
                             -Dquartz=disabled -Dquartz=enabled
     }
 
-    if {${os.major} > 8} {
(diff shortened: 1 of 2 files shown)


barracuda156 (approved):
I cannot test this, but given patch is only used for 10.4, fine with me.

PR #285, opened 2026-09-13
gcc16: remove -no_compact_unwind min os version check on tiger ppc and intel

### LINKED ISSUE #230: GCC16 fails to build on G5 Tiger
G4 is reported working by @ForestExpertise . I have a G4 too but I'm hoping I can fix this faster then that would take to build libgcc16. Intel also obviously works.

Somone on MacRumors has the exact same hardware (2x2ghz G5, 4GB RAM) and noted it works on Leopard. So this is a tiger specific issue. Using make jobs=1 does not fix it either, and we already have a new make involved here. RAM is not exhausted (50% used, about 2GBs)... Opening to track updates. Build log attached (this is the first error I noticed, bootstrap gcc16 stage is crashing with bus_error and compiler_failure, it will completely break when trying to build more later in the build as I've done this multiple times now).

Have a few ideas.

[gcc-error.txt](https://github.com/user-attachments/files/31199195/gcc-error.txt)

alex-free:
The original idea with having a minimum os check for disabling compact_unwind was that the default xcode linkers didn't support it. Since 10.5 and 10.4 both are guaranteed to have a linker that does support it when building gcc16, this is obsolete and actually problematic. 10.6 PPC already uses DWARF (since the original check was not arch specific) instead to encode unwind data for gcc frontend executables (this is again about the actual unwind data encoding of gcc itself, it doesn't affect anything it builds), so there is no need to do a rebuild or revision. My binaries for libgcc/gcc16 also are built like this (as it's actually impossible to do so otherwise in my setups).

GCC developers have known compact_unwind to be too limiting for years to actually encode unwind info while building gcc, which is why this was added in the first place. Although @ForestExpertise is able to build GCC16 on a G4 with tiger with compact_unwind, on Intel it completely fails and on G5 and Rosetta it crashes ld /bintuils, makes the system highly unstable and I've seen it actually KP on both G5 and Rosetta multiple times now (not caused by RAM limit or build jobs either). Even if it doesn't flat out KP, if binutils/ld crash you can't do anything to fix it until you reboot .

After this I realized that if I rebooted, cleaned the build, and started the build with compact unwind disabled for PPC as well (so not gated by arch anymore), it ran all the way through just fine. 

I know we have reports of leopard being OK, but this is just obsolete as the check no longer makes sense and keeps causing problems. I'm going to submit it to upstream(s) like this. It probably makes it faster to build anyways. And for setups that are currently reported as 'ok' (tiger G4, leopard G5) we are just playing with fire as we get gcc16 updated with even more massive code/linking tasks and potentially compact_unwind starts being problematic there too. There's just no benefit to gating this to 10.6 and up anymore, and infact many downsides.

This is after weeks of testing by the way, not some one off. I just hadn't gotten around to submitting it. Having failing builds of gcc16 is not fun on period correct hardware after all.

Closes: https://github.com/macos-powerpc/powerpc-ports/issues/230


lang/gcc16/Portfile
```diff
@@ -276,14 +276,16 @@ platform darwin 8 {
...
-    # DWARF. GCC fixed this in their PR41260 workaround, but made it require 10.6+ minimum. Changes it to no minimum.
-    # Fixes error: ld: 24-bit pointer diff out of range in unwind info in unwind info from a681 and for other executables.
-    if {${configure.build_arch} eq "i386"} {
-        patchfiles-append   darwin8-no_compact_unwind.patch
-    }
+    # the usage of compact unwind. Compact unwind can not be used when building gcc16 due to it's complexity, so fall back to
+    # DWARF. GCC fixed this in their PR41260 workaround, but made it require 10.6+ minimum as the default Tiger and Leopard
+    # linkers did not support it. Changes it to no minimum, as even tiger requires a newer linker that does support the flag.
+    # Fixes error: ld: 24-bit pointer diff out of range in unwind info in unwind info from a681 and for other executables
+    # on Intel. For PowerPC when building through Rosetta or on a G5, this prevents freezing during linking.
```
barracuda156, line 285: While I did not build on 10.5 myself, nobody reported any issues on G5 for 10.5, AFAIR. If there is a bug affecting 10.5, it should be raised to gcc upstream. I am not too eager to add hacks to the primary compiler speculatively.

alex-free: I'll make it 10.4 specific for now, but I wouldn't call it a hack. The min os version check is obsolete as ld can pass the -no_compact_unwind flag by time it gets to this part of the dependency chain for tiger and leopard.

alex-free:
10.4 only for now. I'll let you know what upstream says. And I plan on getting some leopard/snow leo installs going once I get more of my backlog finished with tiger (this is almost certainly broken on intel leopard, I'll try on power too)

PR #289, opened 2026-09-14
openvpn2: add tuntaposx dep for under 10.7, use macports bash shabang on tiger for dns-updown2

alex-free:
Tiger bash is way older then leopard and newer, openvpn2 tries to exec dns-updown2 script to bring up the vpn and fails unless you explicitly point it to macports bash. So add bash as a runtime dep and reinplace shabang to macports bash.

Since 10.6 and below don't have a native tun interface, add that as a runtime dep as well.


net/openvpn2/Portfile
```diff
@@ -33,6 +33,22 @@ depends_lib         port:lz4 \
                     port:lzo2 \
                     path:lib/libssl.dylib:openssl
 
+# Mac OS X 10.7 is the first to have native tun.
```
barracuda156, line 36: 10.6 has `utun` though I am not sure if openvpn can use that.

net/openvpn2/Portfile
```diff
@@ -33,6 +33,22 @@ depends_lib         port:lz4 \
...
+    depends_run port:tuntaposx
+}
+
+# Tiger bash is too old to exec dns-updown2.
+platform darwin 8 {
+    incr revision
+    depends_run-append port:bash
+
+    post-patch {
+        reinplace "s|^#!/bin/bash$|#!${prefix}/bin/bash|" \
```
barracuda156, line 47: I’d drop shebang characters from here and just reinplace `/bin/bash` to `${prefix}/bin/bash`.

barracuda156 (approved):
Not entirely convinced re `tuntaposx` been needed, since 10.6 does have native utun: https://github.com/alexey-lysiuk/macos-sdk/blob/main/MacOSX10.6.sdk/usr/include/net/if_utun.h

But no real hard in runtime dep, so ok.

PR #297, opened 2026-09-18
libvmaf: disable intel asm on 10.6 and lower

alex-free:


multimedia/libvmaf/Portfile
```diff
@@ -40,6 +40,13 @@ configure.args-append   --buildtype release \
                         -Denable_avx512=false \
                         -Denable_tests=false
 
+# The only asm this enables is for avx2.
+if {${os.platform} eq "darwin" && ${os.major} < 11 && \
+${configure.build_arch} in [list i386 x86_64]} {
+    configure.args-append -Denable_asm=false
+    depends_build-delete port:nasm
```
barracuda156: This won’t work (`nasm` is added back below, configure arg should be replaced). I think i386 stuff should also go into one block.

barracuda156 (changes requested):

multimedia/libvmaf/Portfile
```diff
@@ -51,8 +51,12 @@ variant float description "Compile floating-point feature extractors into the li
 }
 
 platform i386 {
-    depends_build-append \
-                        port:nasm
+    # The only asm the default enable_asm=true enables is for avx2.
+    if {${os.platform} eq "darwin" && ${os.major} < 11} {
+        configure.args-append -Denable_asm=false
```
barracuda156:
```suggestion
        configure.args-replace -Denable_asm=true -Denable_asm=false
```

multimedia/libvmaf/Portfile
```diff
@@ -51,8 +51,12 @@ variant float description "Compile floating-point feature extractors into the li
 }
 
 platform i386 {
```
barracuda156:
```suggestion
if {${os.arch} eq "i386" && ${configure.build_arch} ne "ppc"} {
```
Or something to this effect, we do not want Rosetta to have all this.

barracuda156 (changes requested):

multimedia/libvmaf/Portfile
alex-free: Wait, don't we need this to be both x86_64 and i386 (IIRC platform i386 was said to match for both?)

barracuda156 (approved):

PR #300, opened 2026-09-20
xxhash: use gmake port on tiger

alex-free:


devel/xxhash/Portfile
```diff
@@ -0,0 +1,58 @@
...
+                    sha256  5738270935e7c3d38a79b3adf7c9692566ce7895a25f67de43ad52ab504acd32 \
+                    size    1167538
+
+# Avoid build failures due to use of MMX instructions
+# See: https://trac.macports.org/ticket/71687
+compiler.blacklist-append \
+                    {clang < 900}
+
+platform darwin 8 {
+    depends_build-append port:gmake
```
barracuda156, line 26: @alex-free This is kinda getting out of hand. Would you consider updating system `make` instead?
There is a port for that: https://github.com/macos-powerpc/powerpc-ports/blob/main/devel/sl-make/Portfile

barracuda156: Alternatively, just set default `build.cmd` to gmake in the base.

Here: https://github.com/macports/macports-base/blob/3bba6d92e559bca362e6cf2d216333a3bd598401/src/port1.0/portbuild.tcl#L87-L135

alex-free: OK next thing I will focus on. I'm going to figure out something about make (and the incompatible pointer types) and go through the PRs made here (that are just about these things) to remove them all :).

alex-free:
Going to close as I'm about to update my base

